PR #96, opened 2025-12-13
delete network speed detect code

mihneapopesq:
### Task Description

Enhancement:
No need to detect the network speed in seekdb.

The keyword is get_network_speed_from_sysfs in ob_server.cpp.
There's also a timer refresh_network_speed which call this method that can be deleted too.

### Solution Description

deleted the methods and functions that deal with the network speed:
get_network_speed_from_sysfs and refresh_network_speed


hnwyllmm (changes requested):
Thanks for your contribution. I left some comments and please check them.

mihneapopesq:
Initialized properly now

PR #137, opened 2026-01-28
fix: always print malloc sample info regardless of FATAL_ERROR_HANG m…

### LINKED ISSUE #13: We should always print the statistic information of memory allocation
## Summary
We can always print the statisic information of memory allocation now. The action overhead is very light.

## How to fix
In memory_dump.cpp, when FATAL_ERROR_HANG macro is not define, the print_malloc_sample_info() won't be executed.

We can remove the macro define guard.

Image

Zhangg7723:
…acro

Remove the FATAL_ERROR_HANG macro guard around print_malloc_sample_info() so that memory allocation statistics are always printed. The overhead is very light and this information is useful for debugging.

Fixes #13


LINxiansheng (approved):
LGTM

PR #147, opened 2026-01-31
Improve embed error messages

hawkingrei:
## Summary
- prefer user-facing messages from warning buffer and ob_errpkt_str_user_error in embed errors
- include error name and errno in embed exceptions for clarity
- centralize formatting in embed error helper

## Testing
- not run (not requested)

Refs #8

### DIFF
--- a/src/observer/embed/python/ob_embed_impl.cpp
+++ b/src/observer/embed/python/ob_embed_impl.cpp
@@ -300,8 +300,14 @@ ObString format_embed_error(int ret)
 {
   static thread_local char err_buf[1024];
   int64_t pos = 0;
-  ObString errmsg = handle_err_msg(ret);
-  const int err_no = ob_errpkt_errno(ret, false);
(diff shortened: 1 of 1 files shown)


src/observer/embed/python/ob_embed_impl.cpp
```diff
@@ -92,6 +92,9 @@ static pybind11::module builtins = pybind11::module::import("builtins");
 
 #define MPRINT(format, ...) fprintf(stderr, "[seekdb] " format "\n", ##__VA_ARGS__)
 
+std::string handle_err_msg(int ret);
```
LINxiansheng: We can not use STL

hawkingrei:
Updated per request: embed error formatting now uses mysql mapping (ob_mysql_errno_with_check + ob_str_user_error/ob_errpkt_str_user_error), so user sees mysql-oriented message. See latest commit a70cb002.

PR #649, opened 2026-04-22
fix: don't fail trigger SQL prepare on non-anonymous PL symbol lookup

LINxiansheng:
## Issue Description
fix: don't fail trigger SQL prepare on non-anonymous PL symbol lookup

### Solution Description
fix: don't fail trigger SQL prepare on non-anonymous PL symbol lookup
## Passed Regressions
core test
### Release Note


hnwyllmm:
**Comment from code review:**

Yes, you're right, I don't know this part either. I can only use Cursor to communicate with you.

hnwyllmm:
**Comment from code review:**

The comment seems incorrect; there is no anonymous block in lite. In what scenario is `?` not in the secondary namespace?

Issue #868, opened 2026-06-08
Observer crashes with SIGABRT in memory_sanity_abort during optimizer column stat cache access


cms-cms:
## Environment
- **Version**: observer_check_3813692 (OceanBase seekdb)
- **Revision**: 1-2a83673460d32c0573d32f8a16aadccce97f7947
- **Build Branch**: master
- **Build Time**: Jun  3 2026 23:52:09
- **Build Flags**: RelWithDebInfo|Sanity
- **Build Info**: obbuild-sanity-master-101086
- **Is this a temporary version provided by RD?**: No

## Description
The observer process crashed with a `SIGABRT` signal, triggered by an internal memory sanity check failure (`memory_sanity_abort`). The crash occurred while the optimizer was fetching column statistics from the KV cache, specifically within the `ObKVCacheMap::get` call path. This indicates a likely memory corruption issue (e.g., Use-After-Free, out-of-bounds access, or dirty pointer) related to the lifecycle or concurrent reclamation of cache objects. This is not a crash directly caused by a business SQL error code (like `-6005`).

**Key Crash Context:**
- This is an **active abort** (`SIGABRT`), not a regular segmentation fault.
- Crash point: OceanBase's memory consistency check in `memory_sanity_abort` at `./deps/oblib/src/lib/alloc/memory_sanity.cpp:547`.
- The call stack originates from the optimizer's column statistics retrieval path:
  `ObOptColumnStatCache::get_row` -> `ObOptStatService::get_column_stat` -> `ObJoinOrder::compute_table_meta_info`.
- The failure happened during a `ObKVCacheMap::get` operation while reading from the KV cache.

## Steps to Reproduce
**Reproducibility**: Not specified (必现/偶现/不确定/其他).

**Test Changes**: Not specified (增加了xx测试场景/换了新机器/新跑了xx工具/替换新版本后才出的).

**Environment Anomalies/Changes**: Not specified (否/clog盘满过/手工做过。。/注入过xx异常).

**Scenario/Steps**: Please provide detailed steps, workload description, actions taken before the issue, and any test changes. **Strongly discouraged**: only posting ERROR logs or stack traces without context.

**Environment Info**:
- Please provide connection strings for the problematic tenant and the sys tenant, along with deployment information.
- Is the scene preserved? `work 5KwlmAzyUCaft9Hi`

## Log Information
- **Core Dump File**: `/obdata/data/coredump/core-T1_PX_G0-401076-1780875682`
[...]

PR #871, opened 2026-06-10
Fix invalid free of KV cache node

hnwyllmm:
## Task Description
Fix a core dump issue caused by accessing illegal memory when using the kvcache for information collection.

## Solution Description
The issue was an invalid free of a kvcache node, which led to accessing an illegal `mb_handle` and triggered illegal memory access.

During node recycling (retire), the judgment condition included `tenant_id == temp_node->tenant_id_`. The `tenant_id` passed during retire was 0, but the node never had its `tenant_id` set correctly, so this condition was always true.

The fix modifies the node creation to set `tenant_id = sys tenant id`. Additionally, the retire judgment condition is updated to include `tenant_id != 0`.

> For more detailed information, please refer to: https://project.alipay.com/workItem?workItemId=2026060800116606444

## Other Information
- Related link: DIMA-2026060800116606444

## Release Note

### DIFF
--- a/src/share/cache/ob_kvcache_hazard_version.cpp
+++ b/src/share/cache/ob_kvcache_hazard_version.cpp
@@ -96,7 +96,8 @@ void ObKVCacheHazardSlot::retire(const uint64_t version, const uint64_t tenant_i
       while (head != nullptr) {
         temp_node = head;
         head = head->get_next();
-        if (temp_node->get_version() < version || tenant_id == temp_node->tenant_id_) {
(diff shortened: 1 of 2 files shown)


hnwyllmm:
The mapping Dima issue is [[seekdb][ddlmigrate]Core was generated by `/obdata/data/ddl_lite/2/bin/observer ...' Program terminated with signal SIGABRT, Aborted](https://github.com/oceanbase/seekdb/issues/868)

footka (approved):

Issue #867, opened 2026-06-07
Vector APPROXIMATE search fails with -4016 Internal error when the filter column has a prefix-length index


killme2008:
# Vector APPROXIMATE query fails with `-4016 Internal error` when a prefix-length index exists on the filter column

## Environment
- **Version**: `OceanBase 4.3.5.3 seekdb (r1.3.0.0) (Built 100000092026051510-13e493b33b6c3ec22133e5fdf8694e8a4e9bff38 May 15 2026)`
- **Image**: `oceanbase/seekdb:latest` (digest `sha256:e3a46b6520fa…`, the current Docker Hub latest)
- **Reproducibility**: Stable, 100%

## Description
An `APPROXIMATE` (HNSW) vector search that also filters on a scalar column fails with
`ERROR 4016 (HY000): Internal error` **if that filter column has a _prefix-length_ index**
(`CREATE INDEX ... ON t(col(N))`). The same query works when the column has a normal
**full** index, no index at all, or when the prefix index is bypassed with `IGNORE INDEX`.

The trigger is purely the **prefix index** — it does **not** depend on the column length
or the vector dimension (reproduces with `VARCHAR(64)` + a `col(20)` prefix index and
`VECTOR(3)`).

## Minimal reproduction
```sql
CREATE TABLE t (
  id  INT PRIMARY KEY,
  grp VARCHAR(64),
  emb VECTOR(3),
  VECTOR INDEX idx_emb(emb) WITH (distance=l2, type=hnsw, lib=vsag)
);

INSERT INTO t VALUES
  (1,'a','[1,0,0]'), (2,'a','[0,1,0]'), (3,'a','[0,0,1]'),
  (4,'b','[1,1,0]'), (5,'b','[0,1,1]'), (6,'b','[1,0,1]');

-- PREFIX-length index on the filter column
CREATE INDEX idx_grp ON t(grp(20));

-- FAILS:
SELECT id FROM t
WHERE grp = 'a'
ORDER BY l2_distance(emb, '[1,0,0]') APPROXIMATE
LIMIT 3;
-- ERROR 4016 (HY000): Internal error
```

## Expected vs Actual
- **Expected**: the rows of group `a` ordered by approximate distance (here `1, 2, 3`).
- **Actual**: `ERROR 4016 (HY000): Internal error`.

## What works (controls / workarounds)
```sql
-- (1) Full index instead of a prefix index -> OK
CREATE INDEX idx_grp ON t(grp);                 -- works

[...]

PR #873, opened 2026-06-10
Vector APPROXIMATE search fails with -4016 Internal error when the filter column has a prefix-length index

hnwyllmm:
## Task Description
github issue: https://github.com/oceanbase/seekdb/issues/867
Vector APPROXIMATE search fails with -4016 Internal error when the filter column has a prefix-length index.

## Solution Description
In the prefix index scenario, a filter containing a regular field (the `grp` field mentioned in the issue) was directly pushed down to the prefix index table. However, this field does not exist in the prefix index table, causing the plan generated by code generation to attempt to access a non-existent column during execution.
The solution is to remove the push-down filter when the column being pushed down to the index table does not exist. (In the current scenario, the `idx_grp` index will use a range scan instead of pushing down a filter to filter data.)

## Other Information
- DIMA-2026060900116643089

### DIFF
--- a/deps/oblib/src/lib/CMakeLists.txt
+++ b/deps/oblib/src/lib/CMakeLists.txt
@@ -387,6 +387,7 @@ ob_set_subtarget(oblib_lib utility
   utility/ob_backtrace.cpp
   utility/ob_proto_trans_util.cpp
   utility/ob_unify_serialize.cpp
+  utility/ob_rpc_authentication_utility.cpp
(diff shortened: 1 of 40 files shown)


hnwyllmm:
The mapping Dima issue is [fixing the vector search and prefix index pre-filter problem](https://project.alipay.com/workItem?workItemId=2026060900116643089).

footka (approved):

PR #874, opened 2026-06-10
Fix stop logic

hnwyllmm:
## Task Description
Stop operation was incorrectly retrieving the status of another seekdb instance.

## Solution Description
Removed the erroneous code.

## Passed Regressions
core-test

## Other Information
Related internal link: DIMA-2026060800116619443


ep-12221:
The mapping Dima issue is about a problem where after stopping a database deployed via the pkg installer on macOS, starting seekdb using `seekdb-start` from a brew installation causes the monitor deployed by the pkg installer to get stuck in a "stopping" state.

footka (approved):

Issue #693, opened 2026-04-29
Potential memory leak in SqlPx module during DDL migration


cms-cms:
## Environment
(Required) Please provide the output of `bin/observer -V`. If this test also involves a version change for obproxy, please also provide the output of `obproxy -V`.

Mapping relationship for the properties on the right: "Target Branch" and "Iteration":
- **Is this a temporary version provided by RD?** No

## Description
A potential memory leak has been identified in the SqlPx module during DDL migration operations.

## Steps to Reproduce
This issue is consistently reproducible.

## Test Changes
The test scenario was expanded to include additional cases.

hnwyllmm:
PL function execution leads to SQL layer memory leak. In seekdb, when performing parallel DDL to create an index, memory in the `mod=SqlPx` module grew continuously from ~87KB to 4.7GB, eventually causing an OOM crash.

Corresponding SQL: The base table contains a large number of columns, with data for some columns generated by PL functions (`RandomDecimal`, `RandomDate`, `generateRandomTime`, etc.). The parallel DDL operation ran for 6.5 hours before OOM.

Investigation Process:
1.  Identify the leaking module: Logs showed `[MEMORY] hold=...mod=SqlPx`, confirming the SqlPx module's hold value grew linearly from 4KB to 4.7GB, with `count` continuously increasing and `avg_used` stable at ~8KB.

2.  Rule out PL cache factors: Remote environment logs showed PL cache repeatedly failing to add entries (`PLAN_CACHE_CTX_ID hold > limit`), causing the PL function to be recompiled 46,012 times. Initially suspected a leak due to PL cache invalidation. A local control experiment (normal cache vs. repeatedly flushed cache) verified that both groups showed identical SqlPx growth behavior. **PL cache invalidation is not the cause of the leak.**

3.  Narrow down the scope: PX + PL function scenario.
    *   SqlPx growth in parallel SELECT + no PL function: Stable (284KB).
    *   SqlPx growth in parallel SELECT + PL function: Continuous growth (13.7MB+).
    *   Non-parallel SELECT + PL function: SqlPx does not grow (does not take the PX path). Further verification of the non-parallel path: Although SqlPx does not grow, the `SqlExecutor` module also shows continuous growth (from 0 to 36MB in 5 minutes).
    **Conclusion: Repeated calls to PL functions cause continuous memory growth on `exec_ctx.get_allocator`. This is not a PX-specific issue; any long-running SQL that repeatedly calls PL functions will trigger it.**

[...]

PR #880, opened 2026-06-11
PL function has memory leak issue in long running SQL

hnwyllmm:
## Task Description
In long-running SQL statements, frequent calls to PL functions pose a risk of memory leaks.

## Solution Description
1.  **Primary Fix for PL Function Memory Leak:**
    Previously, each call to a PL function could allocate a block of memory using the executor context's allocator. This allocator is an arena allocator, meaning the memory is only freed after the entire SQL statement finishes execution. For SQL statements running for several hours, the memory allocated by PL functions could become substantial.
    The fix changes the allocation to use a temporary allocator specific to the PL function. This memory is released immediately upon completion of each PL function execution.

2.  **Secondary Fix for UDF ParamStore Leak (commit 236ec59):**
    **Symptom:** Similar to the PL `expr_op_ctx` leak, when UDF/PL functions are called repeatedly within a long SQL statement, the `mod=SqlExecutor` memory grows linearly with the number of calls and is only released when the entire SQL statement ends.
    **Root Cause:** `ObExprUDF::ObExprUDFCtx::init_param_store` allocated the UDF's ParamStore (approx. 256B per call) on `exec_ctx.get_allocator()` (the outer SQL's arena). The `build_udf_ctx` function intended to cache the `udf_ctx` via `expr_op_ctx_store_` for reuse. However, the `reset_expr_op` call within the PL function's `final` logic clears this store, causing the cache to be ineffective. This forced a re-initialization of `init_param_store` on every call, leading to repeated allocations on the outer arena that were not freed until SQL completion.
    **Fix:** Added a dedicated `ObArenaAllocator ctx_allocator_` member to `ObExprUDFCtx`. The ParamStore is now allocated on this allocator, binding its lifecycle to the `udf_ctx` itself (freed upon `udf_ctx` destruction) and decoupling it from the outer SQL arena. Since this allocator is a per-instance member, it works correctly for nested/recursive UDFs without requiring backup/restore logic.
    **Changes:** Modified `src/sql/engine/expr/ob_expr_udf.{h,cpp}`, totaling 3 locations and 14 lines added.
    **Verification:** Reproduced locally (`select sum(f_leak(v)) from <large_table>`, with millions of PL calls within a single SQL). Before the fix, `SqlExecutor` memory grew from 16MB to 104MB within 50 seconds. After the fix, memory remained constant at ~2MB (count unchanged) throughout, correctly returned to zero after query completion, and results were correct.

## Other Information
- Related internal link: DIMA-2026042900115822534

## Release Note
Fixed memory leaks in PL functions and UDFs during long-running SQL statements by ensuring temporary memory allocations are freed promptly after each function call, rather than accumulating until the SQL statement finishes.

### DIFF
--- a/src/pl/ob_pl.cpp
+++ b/src/pl/ob_pl.cpp
@@ -2992,17 +2992,10 @@ int ObPLExecState::add_pl_exec_time(int64_t pl_exec_time, bool is_called_from_sq
   return ret;
 }
 
-ObArenaAllocator *ObPLExecCtx::get_top_expr_allocator()
-{
-  return &expr_alloc_;
-}
-
 bool ObPLExecCtx::valid()
 {
   return OB_NOT_NULL(allocator_)
          && OB_NOT_NULL(exec_ctx_)
-         // The interface mapped in through the interface mechanism cannot provide or use the func_ pointer.
-//       && OB_NOT_NULL(func_)
          && OB_NOT_NULL(exec_ctx_->get_sql_ctx())
          && OB_NOT_NULL(exec_ctx_->get_my_session());
 }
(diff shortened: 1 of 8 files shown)


hnwyllmm:
The mapping Dima issue is [[seekdb][ddlmigrate]mod=SqlPx module suspected memory leak](https://github.com/oceanbase/seekdb/issues/693)

footka (approved):

PR #885, opened 2026-06-12
Fix primary-standby HA dag scheduler & fold HA_MID/LOW into HA_HIGH

hnwyllmm:
## Task Description
Fix the primary-standby unavailability issue introduced by commit a25b755dce9 (where standby restore HA dag was rejected with -4007, and the HA scheduler initialization was restored). Additionally, perform memory optimization by folding the three HA priority levels (HA_HIGH, HA_MID, HA_LOW) into HA_HIGH.
Note: The modifications to the HA DAG MAP are not yet complete and can be finalized later as part of code cleanup and optimization efforts.

## Other Information
- Related links:
  - DIMA-2026061100116683586

## Release Note


hnwyllmm:
The mapping Dima issue is that HA is disabled in dagmap, which affects the standby's usability.

footka (approved):

PR #897, opened 2026-06-16
Call auto_increment handler inline to avoid timeout under concurrency

footka:
## Task Description
ObGAISRequestRpc was routing each auto_increment fetch through `ex_rpc::sync_call`. This method dispatches the handler to the tenant ReqWorker pool and blocks the caller, resulting in every fetch pinning 2 workers.

## Solution Description
Modified the logic to call the auto_increment handler inline, avoiding the dispatch through `ex_rpc::sync_call`. This eliminates the need to block the caller and pin an extra worker, preventing potential timeouts under high concurrency.

## Other Information
- Related internal link: DIMA-2026061500116753859

### DIFF
--- a/src/share/ob_gais_rpc.cpp
+++ b/src/share/ob_gais_rpc.cpp
@@ -134,8 +134,6 @@ int ObGAISRequestRpc::next_autoinc_val(const ObAddr &server,
     ret = OB_INVALID_ARGUMENT;
     LOG_WARN("invalid argument", KR(ret), K(server), K(msg));
   } else {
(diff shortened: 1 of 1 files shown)


footka:
The Dima issue mapping is: [[seekdb][bigtrans_lite] Timeout (OB_TRANS_TIMEOUT) occurs when GAIS concurrently fetches auto_increment values for partitioned tables with global indexes, which is a regression compared to the previous version.

hnwyllmm (approved):

hnwyllmm (approved):

PR #899, opened 2026-06-16
fix(seekdb): Remove `#ifdef __ANDROID__` guard around `skip_level3_ccl`

ep-12221:
## Task Description
Remove the platform-specific `#ifdef __ANDROID__` guard from the `skip_level3_ccl` optimization to prevent double incrementing of CCL concurrency counters on non-Android platforms (e.g., macOS). This double counting was causing incorrect CURRENT_CONCURRENCY values in DBA_OB_CCL_RULES and premature blocking of queries with error 11087.

## Solution Description
The `skip_level3_ccl` optimization in `src/sql/ob_sql.cpp` was previously guarded by `#ifdef __ANDROID__` in two locations:
1. Lines 3030-3075 (computing skip_level3_ccl) and 3078-3080 (conditionally skipping the first CCL check in `generate_physical_plan`).
2. Lines 3985-4036 (computing skip_level3_ccl and conditionally skipping the second CCL check in `pc_get_plan`).

On platforms like macOS (where `__ANDROID__` is not defined), both CCL checks ran unconditionally:
- The first check at line 3077 (DATABASE_AND_TABLE scope, after resolve) incremented the concurrency counter.
- The second check at line 4037 (DML + DATABASE_AND_TABLE scope, in the plan cache path) incremented the counter again for DATABASE_AND_TABLE rules.

This fix removes all `#ifdef __ANDROID__` guards, making `skip_level3_ccl` a cross-platform optimization. When an outline's `max_concurrent` is stricter than all DATABASE_AND_TABLE CCL rules, both CCL checks are properly skipped as the outline becomes the binding constraint.

## Passed Regressions
Local verification on macOS (arm64):
- Build succeeded in release mode.
- Test Section 1 (no outline, 2 connections send/reap):
  - conn2 was correctly blocked with error 11087 for `ccl_t1_key_word`.
  - DBA_OB_CCL_RULES showed CURRENT_CONCURRENCY=1 for all 3 rules.
  - conn1 reap succeeded with the correct query result.
- Test Section 2 (with outline max_concurrent=0/3):
  - max_concurrent=0 outline correctly blocked with error 5268.
  - max_concurrent=3 outline: conn2 was blocked by CCL (11087), demonstrating that the outline does not override a stricter CCL rule.
  - Monitoring showed correct CURRENT_CONCURRENCY values.
  - conn1 reap succeeded.

## Other Information
DIMA: 2026051200116053759


ep-12221:
The mapping Dima issue is: "On macOS, when creating SQL concurrency control rules and then running the same SQL from multiple terminals, querying the DBA_OB_CCL_RULES table shows unexpected results."

PR #907, opened 2026-06-17
[CP] fix(semantic_index): reject semantic_distance/semantic_vector_distance in generated column

hnwyllmm:
## Task Description
When creating a vector index on a table containing a generated column with `semantic_distance` or `semantic_vector_distance`, the error changed from `OB_ERR_ADD_INDEX` (-5703) to `OB_INVALID_ARGUMENT` (-4002). Root cause: these two functions were incorrectly marked as valid for generated columns (`VALID_FOR_GENERATED_COL`), but they cannot execute properly in the generated column evaluation context, causing type deduction to fail during index creation.

## Solution Description
Changed the registration flag for `semantic_distance` to `NOT_VALID_FOR_GENERATED_COL`. Overrode `is_valid_for_generated_column` for `semantic_vector_distance` to return false. This ensures that during table creation, the standard error `OB_ERR_ONLY_PURE_FUNC_CANBE_VIRTUAL_COLUMN_EXPRESSION` (-5906 / ERROR 3102) is reported, blocking the issue at the source.

## Passed Regressions
core-test

## Other Information
- Related links: DIMA-2026052800116405537
- Cherry-pick from merge point `14409a0f1d672890be01636351b70ecfb8e49c56` in project oceanbase by ob flow
- Previous code review: [267996](https://code.alibaba-inc.com/oceanbase/oceanbase/pull/267996)
**Warning: This review got code conflicts when auto cherry-picking, please make sure the diff is effective**
- DIMA-2026061600116781015

### DIFF
--- a/src/sql/engine/expr/ob_expr_semantic_distance.cpp
+++ b/src/sql/engine/expr/ob_expr_semantic_distance.cpp
@@ -33,7 +33,7 @@
  {
  
  ObExprSemanticDistance::ObExprSemanticDistance(common::ObIAllocator &alloc)
-     : ObFuncExprOperator(alloc, T_FUN_SYS_SEMANTIC_DISTANCE, N_SEMANTIC_DISTANCE, 2, VALID_FOR_GENERATED_COL, NOT_ROW_DIMENSION)
+     : ObFuncExprOperator(alloc, T_FUN_SYS_SEMANTIC_DISTANCE, N_SEMANTIC_DISTANCE, 2, NOT_VALID_FOR_GENERATED_COL, NOT_ROW_DIMENSION)
  {
  }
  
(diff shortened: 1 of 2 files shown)


hnwyllmm:
The mapping Dima issue is about an error when creating an index on a table with generated columns, reporting ob_error=-4002.

hnwyllmm:
Code Review

**Overall assessment: LGTM**, the logic is correct and addresses the root cause. Details of the review are below.

---

### Analysis of the Fix

The two functions use different approaches, which is reasonable.
| Function | Fix Applied | Reason |
|------|---------|------|
| `semantic_distance` | Constructor flag changed to `NOT_VALID_FOR_GENERATED_COL` | `ObExprSemanticDistance` directly inherits from `ObFuncExprOperator`, so the flag can be passed directly. |
| `semantic_vector_distance` | Override `is_valid_for_generated_column` to return `false` | In the inheritance chain `ObExprSemanticVectorDistance → ObExprVectorDistance → ObExprVector`, the `ObExprVector` constructor hardcodes `VALID_FOR_GENERATED_COL`. Additionally, `is_valid_for_generated_col_` is a `private` member of `ObExprOperator`, so a subclass cannot assign to it directly within its constructor body. Therefore, overriding is the minimally invasive and correct approach in this case. |

---

### Test Coverage

Coverage is sufficient: Both functions are covered for `CREATE TABLE` (virtual t_sd1/t_sd5, stored t_sd2/t_sd6). `semantic_distance` is additionally covered for `ALTER TABLE ADD COLUMN` (t_sd3) and composite expression (t_sd4) scenarios. The `is_valid_for_generated_column` check is independent of the CREATE/ALTER path, so the existing test cases are adequate to verify the blocking behavior.

No changes are needed; this is ready to merge.

footka (approved):

footka (approved):

PR #915, opened 2026-06-18
[seekdb][memtable] Replace global Iterator pool with session-level cache

hnwyllmm:
## Task Description
Replace the global object pool used for allocating memtable range scan iterators (`Iterator<BtreeIterator>`, ~5KB each) with a session-level caching mechanism. The global pool, allocated from the 500 tenant memory, was not observable, never shrank, and did not release memory when idle.

## Solution Description
Introduced a two-path dispatch mechanism to replace the global object pool:
- **Foreground SQL**: Obtains the session via `THIS_WORKER.get_session` and uses a session-level free list (`ObBtreeIterCache`) to cache iterators. Iterators are reused across SQL statements within the same session (zero allocation after prewarm). All memory is released when the session disconnects.
- **Background merge/row estimation**: Session is `nullptr`, uses direct `ob_malloc`/`ob_free`. Memory is released immediately upon task completion.
- **Thread safety**: Only one worker thread executes per session at any given time (PX workers have independent deserialized sessions), so the free list requires no locking.

## Passed Regressions
- `test_query_engine`: PASSED (1/1)
- `test_mvcc_callback`: PASSED (19/19)
- Functional verification: All passed for range scans, point queries, multi-range, scans after DML, transaction scans, reverse scans, LIMIT, scans after freeze+merge, cross-layer scans, concurrent 10 connections, and PX PARALLEL(4).
- sysbench `oltp_read_only` 4 threads 30s: TPS=278, QPS=4444, zero errors.
- sysbench `oltp_point_select` 4 threads 30s: TPS=3519, QPS=3519, zero errors.
- Memory verification: `BtreeIterCache` fully releases after session disconnect, `BtreeIter` is freed immediately after merge completion, and the old `Iterator<BtreeI` tag is completely gone.

## Upgrade Compatibility
No impact on upgrades. This is a pure internal memory management change with no persistent format modifications.

## Other Information
All btree iterator-related memory returns to zero when idle, resolving the issue of permanent peak memory occupancy by the global pool.

Performance Regression:
```
┌──────────────────────┬─────────────┬─────────────┬────────┐
│ Scenario            │ master      │ Current     │ Change │
├──────────────────────┼─────────────┼─────────────┼────────┤
│ point_select        │ 112,973 qps │ 117,432 qps │ +3.9%  │
├──────────────────────┼─────────────┼─────────────┼────────┤
│ oltp_read_write     │ 2,291 tps   │ 2,229 tps   │ -2.7%  │
├──────────────────────┼─────────────┼─────────────┼────────┤
│ select_random_ranges│ 44,357 qps  │ 52,333 qps  │ +18.0% │
└──────────────────────┴─────────────┴─────────────┴────────┘
```
Performance data only confirms no regression.

### DIFF
--- a/src/sql/session/ob_sql_session_info.cpp
+++ b/src/sql/session/ob_sql_session_info.cpp
@@ -17,6 +17,7 @@
 #define USING_LOG_PREFIX SQL_SESSION
 
 #include "ob_sql_session_info.h"
+#include "storage/memtable/mvcc/ob_btree_iter_cache.h"
 #include "rootserver/ob_rs_serial_call.h"
 #include "pl/ob_pl_package.h"
 #include "observer/mysql/obmp_stmt_send_piece_data.h"
@@ -162,6 +163,7 @@ ObSQLSessionInfo::ObSQLSessionInfo(const uint64_t tenant_id) :
       client_non_standard_(false),
       is_session_sync_support_(false),
       job_info_(nullptr),
+      btree_iter_cache_(nullptr),
       failover_mode_(false),
       executing_sql_stat_record_(),
       unit_gc_min_sup_proxy_version_(0),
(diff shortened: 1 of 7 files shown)


hnwyllmm:
The mapping Dima issue is about optimizing memory usage for Iterator/BtreeI labels.

footka:
> **src/sql/session/ob_sql_session_info.h:80**

Does px seekdb still mock sessions?

hnwyllmm:
> **src/sql/session/ob_sql_session_info.h:80**

Why is it called a mock session? Isn't it just serializing and deserializing the session information once?

footka (approved):

PR #916, opened 2026-06-18
Fix macOS compilation and remove io_submit health check

ep-12221:
## Task Description
Fix macOS compilation and remove io_submit health check.

## Solution Description
Arrow depends on snappy, but on macOS, it is not automatically linked. Added manual linking.
Removed the io_submit health check, as it may block.

## Passed Regressions
core-test

## Release Note


ep-12221:
The mapping Dima issue is about removing io_submit on Mac.

PR #917, opened 2026-06-22
fix(seekdb): Fix Windows SDK `OPTIONAL` macro naming conflict with C++ enum value

ep-12221:
## Task Description
Fix a compilation error on Windows caused by a naming conflict between a Windows SDK macro and a C++ enumerator.

## Solution Description
**Root Cause**: The Windows SDK header file `winnt.h` defines an empty macro `OPTIONAL` for SAL annotations, such as `_In_opt_`. When the C++ code in `ob_req_packet_code.h` uses `OPTIONAL` as an enumerator in `enum class ObReqCheckSumCheckLevel`, the MSVC/clang-cl preprocessor expands it into an empty string.
**Error File**: `deps/oblib/src/rpc/frame/ob_req_packet_code.h:102`
**Error Line**
```cpp
enum class ObReqCheckSumCheckLevel
{
INVALID
FORCE
OPTIONAL, // ← Expanded to empty by the macro, causing a syntax error
DISABLE
};
```
**Impact Scope**: All compilation units that directly or indirectly `#include` `ob_req_packet_code.h` (approximately 200+ `.cpp` files). This only occurs when building on Windows; Linux/macOS are not affected.

## Passed Regressions
- **Original failing file compilation**: `ob_geo_func_covered_by2.cpp.obj` and 15+ related geo files in the same include chain compiled successfully, with no further `OPTIONAL`-related syntax errors.
- **Macro conflict verification**: After `#undef OPTIONAL`, the enum expands correctly, and references to `ObReqCheckSumCheckLevel::OPTIONAL` resolve normally.
- **Linked library correction**: The removed `zstd_objs.lib` and `lz4-all.lib` entries in `CMakeLists.txt` have been confirmed to be covered by `zstd_1_3_8_objs.lib`; no additional linking is required.
- Note: proto file generation (`protoc.exe`) cannot be executed due to CI sandbox restrictions. This is an existing environment issue and was not introduced by this change.

## Other Information
DIMA: 2026061800116825248


ep-12221:
The mapping Dima issue is [[windows] Windows compilation failure]

hnwyllmm (approved):

PR #918, opened 2026-06-22
fix(vector-index): resolve CREATE VECTOR INDEX error 5703 (two root causes from RPC removal 7321e32e)

footka:
## Task Description
Fix the issue where CREATE VECTOR INDEX fails with error 5703, which was caused by two root issues introduced by the RPC removal in commit 7321e32e.

## Solution Description
1. **index_schema_ UAF**: The `ObCreateIndexArg.index_schema_` used the lazy request-arena allocator, which was freed when the CREATE-INDEX request ended while the async DDL task was still running. This was fixed by binding `index_schema_` to the arg-owned `allocator_`.
2. **async_call<Res> dropped the result**: The `async_call_impl` function filled a local `Res` object but never copied it to `handle->result_`. Consequently, the Query Coordinator (QC) always read a default-constructed `ObPxRpcInitSqcResponse` (with `rc_=OB_NOT_INIT`) and incorrectly judged the SQC initialization as failed, leading to error -5703 under `parallel(18)`. The fix involves:
   - Having `work(h->result)` fill the handle directly.
   - Making `done_` a release/acquire `std::atomic`.
   - Storing handles in `ObSEArray` via a small `HandleRef` wrapper instead of `std::vector`.

## Other Information
- Related internal link: DIMA-2026061600116774641

### DIFF
--- a/src/observer/net/ob_ingress_bw_alloc_service.cpp
+++ b/src/observer/net/ob_ingress_bw_alloc_service.cpp
@@ -118,20 +118,31 @@ int ObNetEndpointIngressManager::collect_predict_bw(ObNetEndpointKVArray &update
       }
     }
   }
-  // single-replica: every endpoint is local; dispatch directly to the local
-  // network frame instead of the removed ObNetEndpointPredictIngressProxy.
-  for (int64_t i = 0; OB_SUCC(ret) && i < update_kvs.count(); i++) {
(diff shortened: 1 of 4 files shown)


footka:
The mapping Dima issue is about fixing more asynchronous RPC semantic implementation errors.

hnwyllmm (approved):

PR #923, opened 2026-06-23
fix(px): own SQC serialize buffer to avoid UAF

footka:
## Task Description
Fix a use-after-free (UAF) issue related to the serialization buffer used by the SQC (Sub Query Coordinator).

## Solution Description
The fix ensures the SQC owns its serialization buffer, preventing it from being freed while still in use.

## Other Information
Related internal issue: DIMA-2026062200116877396


footka:
The mapping Dima issue is: [seekdb][core][SIGSEGV] Crash in ObExecContext::get_my_session — PX dispatch_sqcs asynchronous callback path.

hnwyllmm:
> **src/sql/engine/px/ob_dfo_scheduler.cpp:495**

Using shared_ptr this way allocates memory twice: once for char and once for shared_ptr.
1. Using make_shared can allocate memory only once.
2. Consider using unique_ptr instead of shared_ptr.

hnwyllmm (approved):

PR #928, opened 2026-06-25
[seekdb][memory] Reduce glibc and CoStack fragmentation

hnwyllmm:
## Task Description
This change reduces unnecessary memory allocation pressure and background work during startup and idle periods in seekdb, which is being optimized as a lighter single-process database. The specific issues addressed are:
- Unnecessary gRPC service registration in embed mode, which could initialize protobuf/gRPC-related memory.
- Inefficient parser non-reserved keyword trie construction using many separate same-sized node allocations.
- Default thread stack sizing using an outdated 512KB-era initialization path instead of the current 256KB target.
- Tenant unit refresh persisting metadata even when the computed unit config did not change.

## Solution Description
The final diff includes the following changes to reduce startup/idle memory usage:
- Skip gRPC service registration in embed mode via `lib::is_embed_mode` to avoid unnecessary gRPC/protobuf initialization.
- Align default thread stack sizing to 256KB during process initialization and deployment templates.
- Replace the misleading `SIG_STACK_SIZE` deduction with `THREAD_STACK_RESERVED_SIZE`; no alternate signal stack is installed here.
- Build the parser non-reserved keyword trie from a counted contiguous node pool instead of many small per-node allocations.
- Avoid persisting tenant unit metadata when the computed unit config does not actually change.

The protected stack allocator remains on the original OB allocator path. An experiment using `mmap/munmap` to create thread stack memory (which saved ~15M memory but caused a 7% sysbench performance regression) was reverted and is not part of the final MR diff.

## Passed Regressions
Build validation after the CoStack `mmap/munmap` revert:
```bash
cd build_debug && ob-make
```
Result: passed, return code `0`.

Additional checks:
```bash
git diff --check
```
Result: passed.

## Upgrade Compatibility
No on-disk format, protocol, or SQL compatibility change is expected. The retained changes are startup initialization, allocator-layout, and metadata persistence-frequency optimizations.

## Other Information
The final MR diff intentionally excludes temporary investigation artifacts and reverted experiments:
- Temporary GLIBC/CoStack allocation diagnostics.
- `docs/developer-guide/zh/glibc-memory-fragment-notes.md`.
- CSFetcher `change_stream_min_dep_lsn` interval/cache changes.
- `MODIFY_GC_SNAPSHOT_INTERVAL` and snapshot_gc_scn persistence-path changes.
- CoStack protected stack `mmap/munmap` allocation.
- Tenant freeze-info reload interval change.

Historical commit messages still show these investigation steps, but the final diff is the authoritative review scope.

## Release Note
No user-visible workaround. These changes reduce unnecessary startup/idle memory work in seekdb, especially embedded startup paths and parser trie initialization.


hnwyllmm:
The mapping Dima issue is about optimizing glibc memory.

footka (approved):

PR #929, opened 2026-06-25
[seekdb][storage] Reduce SHARE_BLOCK resident memory

hnwyllmm:
## Task Description
`ObTenantStorageMetaService` kept two `ObSharedObjectReaderWriter` instances for tablet/sstable metadata persistence. Each writer held a 2MB `SHARE_BLOCK` buffer for the tenant lifetime, so `shared_object_rwriter_` and `shared_object_raw_rwriter_` contributed resident memory even when no metadata write was in progress.

## Solution Description
Introduce a per-call write session inside `ObSharedObjectReaderWriter` and move the 2MB `SHARE_BLOCK` buffer lifetime into that session. The writer object now keeps only the shared macro append state, while raw, batch, and linked writes allocate the 2MB buffer on demand, flush through the existing shared object layout, and release the buffer when the write call exits.
The change preserves existing address semantics and disk layout: `BLOCK` and `RAW_BLOCK` remain unchanged, shared macro blocks continue appending by `offset_`, and block switching still happens only when the current macro block cannot fit the next logical object. A single object that cannot fit into an empty shared macro block is rejected with `OB_NOT_SUPPORTED`.

## Passed Regressions
- `git diff --check`
- `ob-make --inc -s src/storage/blockstore/ob_shared_object_reader_writer.cpp -j 8`
- `ob-make -j8 test_sstable_meta`
- `./unittest/storage/blocksstable/test_sstable_meta`
- `ob-make -j8 test_tablet`
- `./unittest/storage/test_tablet`
- `ob-make -j8 test_linked_macro_block`
- `./unittest/storage/test_linked_macro_block`
- Manual debug process check: after startup and a metadata workload, `oceanbase.v$ob_memory` reported `SHARE_BLOCK` hold/used/count as `0/0/0`.

## Upgrade Compatibility
No data format or address format change. The change only adjusts in-memory buffer ownership during metadata write calls.

## Other Information
This MR includes two commits: an initial raw-writer optimization and a follow-up refactor that unifies both shared object writers under the session-based temporary buffer model.

### DIFF
--- a/src/storage/blockstore/ob_shared_object_reader_writer.cpp
+++ b/src/storage/blockstore/ob_shared_object_reader_writer.cpp
@@ -706,12 +706,51 @@ const char *ObSharedObjectIOCallback::get_data()
 
 //=================================== ObSharedObjectReaderWriter =============================
 const MacroBlockId ObSharedObjectHeader::DEFAULT_MACRO_ID(0, MacroBlockId::AUTONOMIC_BLOCK_INDEX, 0);
+
+struct ObSharedObjectReaderWriter::ObSharedObjectWriteSession final
+{
+public:
+  ObSharedObjectWriteSession()
+    : data_("SHARE_BLOCK", 0, false, false/*use_fixed_blk*/),
+      clean_object_handle_(), clean_offset_(0), clean_align_offset_(0),
+      clean_hanging_(false), has_clean_state_(false)
+  {}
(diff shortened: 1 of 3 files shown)


hnwyllmm:
The mapping Dima issue is about optimizing the SHARE_BLOCK memory module.

footka (approved):

PR #942, opened 2026-06-26
[seekdb][pl] Fix DBMS index refresh in embedded mode

hnwyllmm:
## Task Description
In an embedded build created with `-DBUILD_EMBED_MODE=ON`, executing `CALL dbms_index_manager.refresh;` failed with `ERROR 1235 (0A000): Not supported feature or function`.
The package SQL and native implementation were both present, but embedded PL compilation had a separate direct-dispatch path for `PRAGMA INTERFACE` routines. That path skipped the normal interpreter metadata setup used by non-embedded builds.

## Solution Description
Remove the embedded-only PL interface direct-dispatch branch and package codegen skip branch.
Embedded and non-embedded builds now use the same PL compile path: prepare expressions, retain the AST, and let the interpreter execute `PL_INTERFACE -> spi_interface_impl`.

## Passed Regressions
- `ob-make -j4 ob_pl` in `build_release_embedded`
- `ob-make -j4 seekdb` in `build_release_embedded`
- Started the rebuilt embedded `seekdb` on a temporary local port and verified `CALL dbms_index_manager.refresh;` succeeds.
- Server log confirmed `ObDBMSIndexManager::refresh` ran and `change stream refresh scn caught up`.

## Upgrade Compatibility
No schema or data format change. This only aligns embedded PL interface routine compilation with the existing non-embedded path.

## Other Information
MR range contains one commit:
- `ebebc2f1787 fix pl calling in embed mode`

### DIFF
--- a/deps/oblib/src/lib/signal/ob_signal_handlers.cpp
+++ b/deps/oblib/src/lib/signal/ob_signal_handlers.cpp
@@ -206,11 +206,9 @@ void coredump_cb(volatile int sig, volatile int sig_code, void* volatile sig_add
     int64_t ip = con->uc_mcontext.regs[30];
     int64_t bp = con->uc_mcontext.regs[29];
 #endif
-    #ifndef OB_BUILD_EMBED_MODE
-      void* addrs[64];
-      int n_addr = light_backtrace(addrs, ARRAYSIZEOF(addrs), bp);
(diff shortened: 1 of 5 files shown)


hnwyllmm:
The mapping Dima issue is that the embedded mode dbms_index_manager.refresh cannot be used.

footka (approved):

PR #947, opened 2026-06-29
Fix PX-budget hold-and-wait deadlock: admit PX just before execute_plan

footka:
## Task Description
Move PX admission from `open_plan` to `do_open_plan` right before `execute_plan` (after open-phase pre-execution work like direct-load's `create_hidden_table`), for all plans.

## Solution Description
Previously PX was reserved at `open_plan` and held idle through `create_hidden_table`'s same-table DDL conflict retry, starving the conflicting index build that also needs PX -- a circular hold-and-wait. Admitting right before parallel execution keeps the reservation tight to actual use; the `admission_acquired` guard keeps the admit idempotent across the txn-set-violation retry loop; release is unchanged.

## Other Information
- Related internal issue: DIMA-2026062500116973490

### DIFF
--- a/src/sql/ob_result_set.cpp
+++ b/src/sql/ob_result_set.cpp
@@ -104,15 +104,10 @@ OB_INLINE int ObResultSet::open_plan()
     LOG_ERROR("invalid physical plan", K(physical_plan_));
   } else {
     has_top_limit_ = physical_plan_->has_top_limit();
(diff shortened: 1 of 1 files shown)


footka:
The Dima mapping issue is: [[SeekDB][ddlmigrate] DDL hangs for a long time during stress testing, causing the test to fail to exit because the `copy_table_dependent_objects` subtask cannot complete.

hnwyllmm (approved):

PR #958, opened 2026-07-01
[storage] remove checkpoint diagnose residue

hnwyllmm:
## Task Description
The `checkpoint diagnose` feature's main path has been removed, but the repository still contains residual components: an independent manager, tenant lifecycle integration, checkpoint/memtable/MDS diagnose instrumentation points, and the integer `trace_id` passing chain used for this path. These continue to retain `CkptDgnMem` / `CkptDgnMemCU` related memory tags and invalid code paths.

## Solution Description
- Delete the `ObCheckpointDiagnoseMgr` module, its compilation entry point, tenant module lifecycle integration, and `_checkpoint_diagnose_preservation_count`.
- Refactor the checkpoint / freezer / memtable / lock memtable / MDS flush interfaces to remove the private integer `trace_id` passing chain for checkpoint diagnose.
- Delete the diagnose helper macros, diagnose fields, and reporting logic on memtable/MDS.
- Clean up virtual table comments, mysql test results, and residual log fields from the compilation.

## Passed Regressions
- `cmake --build . --target seekdb -j4`
- `strings build_debug/src/observer/seekdb | rg "CkptDgnMem|CkptDgnMemCU|CkptDgnMemNode|CkptDgnCUNode|checkpoint_diagnose|ObCheckpointDiagnose"` returns no hits.
- `rg -n "CkptDgnMem|CkptDgnMemCU|CkptDgnMemNode|CkptDgnCUNode|checkpoint_diagnose|ObCheckpointDiagnose" src tools` returns no hits.
- `git diff --check`
- Verified that the foreground process can start, initialize, and write `seekdb.log`. The exit reason was a socket creation failure in `ob_expr_uuid.cpp` (`ret=-4016 errno=1`) due to the sandbox environment, which is not a checkpoint diagnose issue introduced by this change.

## Upgrade Compatibility
Only removes residual code from the already decommissioned checkpoint diagnose feature. Does not introduce new upgrade compatibility requirements.

## Other Information
- This MR contains only 1 commit: `17e26050e17 fix(storage): remove checkpoint diagnose residue`
- Related internal link: DIMA-2026063000117090987


hnwyllmm:
The mapping Dima issue is [CkptDgnMem memory analysis](

footka (approved):

PR #959, opened 2026-07-01
Add LLVM Source-based Code Coverage Support

hnwyllmm:
## Task Description
Add LLVM source-based code coverage support in seekdb.

## Other Information
- Related internal link: DIMA-2026062300116930662

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -26,6 +26,8 @@ ANDROID_BUILD=false
 LLD_OPTION=ON
 ASAN_OPTION=ON
 STATIC_LINK_LGPL_DEPS_OPTION=ON
+ENABLE_BOLT_OPTION=ON
+WITH_COVERAGE=OFF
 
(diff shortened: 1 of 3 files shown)


hnwyllmm:
The mapping Dima issue is about integrating clang source-based code coverage into seekdb.

footka:
> **src/observer/main.cpp:30**

Why not just call __llvm_profile_set_filename(base_dir) directly?

hnwyllmm:
> **src/observer/main.cpp:30**

Calling `__llvm_profile_set_filename` and setting the `LLVM_PROFILE_FILE` environment variable achieve the same end result. `__llvm_profile_set_filename` directly sets the output path for the profraw file, whereas using `LLVM_PROFILE_FILE` means the profile runtime initializer reads this environment variable to set the path later during initialization. The code has been changed to use `__llvm_profile_set_filename`.

footka (approved):

PR #960, opened 2026-07-01
Optimize log to reduce binary size

LINxiansheng:
## Task Description
optimize log to reduce binary size

## Passed Regressions
core test


LINxiansheng:
The mapping Dima issue is [log optimization](

PR #973, opened 2026-07-06
Skip disk health detection on macOS

LINxiansheng:
## Task Description
Skip disk health detection on macOS

## Solution Description
Skip IO fault detector tasks on macOS and keep device health status normal to avoid OB_DISK_HUNG false positives during health checks

## Passed Regressions
core test

## Release Note


LINxiansheng:
The mapping Dima issue is [Fix the timeout issue caused by io_submit blocking on macOS](

PR #979, opened 2026-07-06
Optimize fulltext index building performance

LINxiansheng:
## Task Description
optimize fulltext index building performance

## Solution Description
1. optimize fulltext index
2. reuse parser state for FTS index row generation

## Passed Regressions
core test


LINxiansheng:
The mapping Dima issue is [optimize fulltext index building performance](

PR #1000, opened 2026-07-09
[seekdb][vector index] Fix schema mgr GC and validate VSAG params

hnwyllmm:
## Task Description
This MR addresses two main issues:
1.  An `invalid argument` error was raised during async vector index creation in testing, but the error should be reported to the user at the time of index creation (DDL).
2.  A SchemaMgr parallel issue where Garbage Collection (GC) could free a schema manager object while it was still being built, leading to a sanity abort.

The specific trigger was a coredump (DIMA-2026070600117218556) during async vector index access. This occurred after DDL accepted an invalid HNSW_SQ parameter combination, which later caused VSAG incremental index initialization to fail. During the subsequent schema fallback path, schema mgr GC could free a schema manager that was either created but not yet inserted into the cache, or removed from the cache but not yet fully released. This left a pinned schema mgr object in an unsafe state, triggering the abort.

## Solution Description
1.  **Harden schema mgr cache elimination ownership:** Modified the GC fallback path to not free a schema mgr pointer if it cannot be found in the cache. This is because the pointer may be in a constructing state (before cache insertion) or already removed from the cache and awaiting release by its owning path.
2.  **Add VSAG create-index validation during DDL:** Introduced a validation path for VSAG index creation that is called during DDL vector index parameter checking. This validation reuses the same VSAG JSON construction and `max_degree` adjustment logic as the runtime `create_index` but does not retain an index handler.
3.  **Validate parameter shapes for HNSW_SQ and HNSW_BQ:** The validation now checks both snapshot and incremental initialization parameter shapes. Since the incremental path converts SQ/BQ to HGRAPH, DDL will now reject invalid combinations (e.g., `m=22` and `ef_construction=30`) before async index initialization begins.
4.  **Preserve and expose VSAG error messages:** Error messages from `Factory::CreateIndex` are now preserved, and parameter validation failures are exposed as user-visible DDL errors.

## Passed Regressions
- `git diff --check`
- `git diff --cached --check`
- Source-level validation against the coredump path and the VSAG async initialization path.

A full build/regression suite was not executed in this worktree because no build directory/Makefile was initialized; `ob-make -n observer` reported no `observer` target.

## Upgrade Compatibility
No impact on storage format or upgrade compatibility. The changes only cause invalid DDL to be rejected earlier and tighten the ownership rules for schema mgr GC.

## Other Information
This MR contains two commits:
- `787ef25eaac` fix schema mgr cache elimination ownership
- `e63b6f9f45e` validate vsag index params during ddl

Local untracked files `dima.txt` and `2` are analysis artifacts and are not included in the commit.

## Release Note
- **Bug Fix:** Fixed a potential coredump during async vector index creation by hardening schema manager garbage collection logic to prevent freeing objects that are still under construction or pending release.
- **Enhancement:** Added validation for VSAG (HNSW/HGRAPH) index parameters during DDL execution. Invalid parameter combinations (e.g., `ef_construction` smaller than the effective HGRAPH `max_degree`) are now rejected at the time of index creation with a clear error message, rather than failing later during asynchronous initialization.


hnwyllmm:
The mapping Dima issue is [[seekdb][stress_duomo_lite] A core dump occurred during execution, core-T1_CSWorker220-2699203-1783320782](https://github.com/oceanbase/oceanbase/issues/xxxxx).

footka (approved):

PR #1001, opened 2026-07-09
Remove log_cold_cache

footka:
## Task Description

log_cold_cache is a secondary in-memory cache for PALF history logs. It reduces repeated disk reads after logs have moved out of group_buffer; it is a performance optimization, not a correctness dependency.

For seekdb, removing it is acceptable: recent logs are still served from group_buffer, and older logs can still be read from PALF log files. Even if a standby fetches logs that have moved out of group_buffer, the impact is only an extra disk read, not correctness risk.

Experiments show no deterministic IVF/HNSW query latency regression after removal. Meanwhile, **it eliminates log_kv_cache peaks during insert/index build: about 85 MiB for IVF and 73 MiB for HNSW.**

## Release Note


footka:
The mapping Dima issue is about removing cold_cache from clog logs.

hnwyllmm (approved):

PR #1002, opened 2026-07-09
[seekdb][pl] Avoid static SQL buffer leak

hnwyllmm:
## Task Description
High-concurrency DBMS_VECTOR.refresh_index_inner refresh/DDL workload reported PL allocator leak diagnostics (e.g., `memory leak(old_blk->ref_=48)`). A previous fix for CALL/EXECUTE temporary argument arrays is already in the branch. This new occurrence hits another static SQL path: the tree-walk PL interpreter copied embedded SQL text from `ObPLSqlStmt::get_sql`/`get_ps_sql` into `ctx->allocator_` as null-terminated `const char *` buffers. These SQL strings are statement-scope temporaries, but `ctx->allocator_` is backed by `ObPLAllocator1`/`ObVSliceAlloc` and is checked at PL execution teardown, so repeated static SQL execution could leave cached slice references reported as leaks.

## Solution Description
Pass static PL SQL text through SPI as `ObString` instead of copying it to `const char *` buffers in `exec_sql`.
Specifically:
- Change `spi_query_into_expr_idx`, `spi_execute_with_expr_idx`, `spi_query`, `spi_execute`, `spi_inner_execute` to accept `ObString` SQL text.
- Pass `ObPLSqlStmt::get_sql`/`get_ps_sql` directly from the tree-walk interpreter.
- Pass `ObPLSqlInfo::sql_`/`ps_sql_` directly from the simple_execute path.
- Keep `dup_cstr` only for cursor open paths where C-string SQL is still persisted by cursor APIs.
This removes the redundant SQL text allocation from the PL allocator while keeping cursor SQL lifetime handling unchanged.

## Passed Regressions
- `git diff --check`
- `git diff --cached --check`
- `build_release: ob-make observer`

## Upgrade Compatibility
No upgrade compatibility impact. This changes internal SPI/static PL SQL parameter passing from `const char *` to `ObString` and does not change SQL semantics or persisted metadata.

## Other Information
MR contains one commit: `24d2e8b82e1 fix(pl): avoid static sql buffer leak`
Untracked local files 2 and nohup.out are not included.
Related internal link: DIMA-2026062500116992865

### DIFF
--- a/src/pl/ob_pl.cpp
+++ b/src/pl/ob_pl.cpp
@@ -3992,7 +3992,7 @@ int ObPL::simple_execute(ObPLExecCtx *ctx, int64_t argc, int64_t *argv)
       if (OB_FAIL(ret)) {
       } else if (sql_info.params_.empty()) {
         OZ (ObSPIService::spi_query(
-          ctx, sql_infos.at(i).sql_.ptr(), sql_infos.at(i).stmt_type_,
+          ctx, sql_infos.at(i).sql_, sql_infos.at(i).stmt_type_,
           sql_info.into_.get_data(), sql_info.into_.count(),
           sql_info.data_type_.get_data(), sql_info.data_type_.count(),
           sql_info.not_null_flags_.get_data(),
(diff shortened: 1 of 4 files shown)


hnwyllmm:
The Dima mapping issue is [[seekdb][ddltest_duomo][errcode=-4016] memory leak (old_blk->ref_=400), reference count leak under high-concurrency refresh/DDL operations.

footka (approved):

PR #1003, opened 2026-07-09
Remove legacy OceanBase protocol support

footka:
## Task Description
Keep only the capabilities required by the standard MySQL protocol, and drop OceanBase private protocol extensions for proxy coordination, driver identification, extra info transmission, private prepared-statement optimization commands, and private session-state propagation.

## Release Note


footka:
The mapping Dima issue is about removing the ob20 protocol.

hnwyllmm (approved):

PR #1004, opened 2026-07-09
[seekdb][allocator] Move chunk cache into block set

hnwyllmm:
## Task Description
AChunkMgr currently keeps a global free chunk cache. In idle periods these cached chunks are not released, which can leave tens of MB of memory resident even after the workload has stopped. After BlockSet gained purge/wash capability, the chunk cache can be managed closer to block ownership so idle whole-free chunks can be reclaimed by BlockSet wash.

## Solution Description
Move normal chunk reuse from AChunkMgr/ObTenantCtxAllocator into BlockSet.
- BlockSet now allocates chunks through ObTenantCtxAllocator and keeps whole-free reusable chunks in its free-list.
- When a whole-free chunk expires during BlockSet wash, BlockSet directly releases the chunk through the tenant ctx allocator.
- Page-level washed blocks are still not reused; when the containing chunk becomes fully free, the chunk is directly unmapped instead of being cached.
- AChunkMgr no longer keeps a global normal chunk free-list cache, and the reload path no longer applies memory_chunk_cache_size to AChunkMgr.
- Remove the old common_realloc global sync_wash retry path.

## Passed Regressions
- git diff --check
- git diff --cached --check
- ob-make test_chunk_mgr test_tenant_ctx_allocator test_block_set
- deps/oblib/unittest/lib/test_block_set
- deps/oblib/unittest/lib/test_tenant_ctx_allocator
- deps/oblib/unittest/lib/test_chunk_mgr
- ob-make observer

## Upgrade Compatibility
No on-disk or protocol compatibility impact. This changes in-process allocator cache ownership and reclaim behavior.

## Other Information
This MR contains one commit:
- 7e12e67dcfb move chunk cache into block set

Untracked local files are intentionally not included:
- 2
- docs/design/


hnwyllmm:
The mapping Dima issue is about analyzing the chunk manager.

hnwyllmm:
> **deps/oblib/src/lib/alloc/block_set.cpp:398**

Why is a new `reclaimed_size` variable being added here? I see that `washed_size` is no longer used later on either. It seems like adding this variable doesn't serve a purpose?

hnwyllmm:
> **deps/oblib/src/lib/alloc/block_set.cpp:430**

Why must it be a chunk with can_cache_whole_chunk to be released? In theory, any complete chunk should be releasable regardless of the condition, right?

hnwyllmm:
> **deps/oblib/src/lib/alloc/block_set.cpp:207**

Let's change it so that regardless of whether it's washed or not, we put it into the free_list_.

hnwyllmm:
> **deps/oblib/src/lib/alloc/block_set.h:55**

Since we're not doing anything with it, let's just delete the code.

hnwyllmm:
> **deps/oblib/src/lib/alloc/ob_tenant_ctx_allocator.cpp:238**

Why hasn't AChunkMgr been deleted? Should we delete it?

PR #1032, opened 2026-07-13
Remove Lua diagnose execution

footka:


footka:
The mapping Dima issue is about removing Lua diagnostics.

hnwyllmm (approved):

PR #1038, opened 2026-07-13
[seekdb][memory] Reduce small-object memory overhead

hnwyllmm:
## Task Description
SeekDB lite runs in a single-node, single-tenant, single-LS deployment model, but several inherited OceanBase modules still keep data structures sized for distributed or high-concurrency workloads. These small-object allocations are individually modest, but they become always-resident startup/idle overhead when many hash buckets, queues, pools, and diagnostic collectors are initialized eagerly.

MemDumpContext is intentionally excluded from this MR because it is handled separately.

## Solution Description
- Remove unused query response time statistics
  - delete tenant response-time collector initialization and SQL-path collection;
  - remove `SHOW QUERY_RESPONSE_TIME`;
  - remove `oceanbase.__all_virtual_query_response_time`, `information_schema.QUERY_RESPONSE_TIME`, and `V$OB_QUERY_RESPONSE_TIME_HISTOGRAM`;
  - remove the corresponding unittest and CMake registration.
- Simplify low-cardinality DDL bookkeeping
  - replace DDL task heartbeat hash maps and bucket locks with small arrays protected by a spin lock;
  - remove the global DDL merge tablet bucket lock and guard task. In seekdb-lite, direct-load DDL merge is single-table scoped and the DAG/tablet dependency already provides the needed exclusion.
  - Create major-merge progress maps only for an active merge round, then destroy them in `clear_cached_info` instead of keeping the maps resident while idle.
- Simplify rarely populated runtime state
  - replace `ObLongopsMgr`'s bucket map and pre-initialized concurrent FIFO allocator with a small array and on-demand `ob_malloc` stat nodes;
  - remove `ObTabletStreamPool`'s fixed 100-node free-list cache and allocate stream nodes on demand while preserving the existing LRU retirement and upper-limit behavior.
- Reduce seekdb-lite object-pool/cache footprint
  - cap lite-mode memtable object pool at 256 instead of deriving 5000 from the tablet pool size;
  - reduce MDS / tx-data allocator concurrency from 32 to 8;
  - reduce the single-LS transaction context map bucket count from 16384 to 1024;
  - set `ParamCharsType` SEArray block size to 256;
  - remove unused Oracle NUMBER cast boundary precomputed arrays.
- Simplify local GTI source
  - remove the 16-entry trans-id prefetch cache and adaptive preallocation counters;
  - refill a local trans-id range from `ObTransIDService` on demand.

**Memory Impact**
For seekdb-lite single-tenant/single-LS
- Conservative startup/idle resident-memory reduction: about 2.0-2.2 MiB.
- Runtime retained-memory upper-bound reduction after workload churn: 10 MiB+.

Major contributors
- memtable object-pool cap: 5000 -> 256 cached objects, about 7.7 MiB by object-size estimate;
- query response time collector removal: about 0.4 MiB per tenant and removes SQL-path atomic accounting;
- DDL heartbeat, Longops, TabletStreamPool, and major-merge progress structures no longer reserve large bucket arrays or fixed node caches while idle;
- `ParamCharsType` saves roughly 7.5 KiB per affected cached plan compared with the default 8 KiB block allocation.

The estimate excludes MemDumpContext and avoids counting optimization attempts that were later reverted or already absorbed by master.

## Passed Regressions
- Refreshed with latest `origin/master` (`97bc95c1d5c`) and resolved the local merge.
- `build_release: ob-make observer` passed.

**Performance Comparison**
Dedicated cloud-host ABBA sysbench comparison against fresh single-node seekdb instances
- baseline: `origin/master` at test baseline (`97bc95c1d5c`)
- branch: current MR branch after master refresh (`d57b22d0c1a`)
- host: Alibaba Cloud Linux 3 VM, 8 vCPU, 14 GiB memory
- server config: `memory_limit=2G`, `datafile_size=2G`, `datafile_maxsize=4G`, `log_disk_size=2G`
- sysbench: 1.0.20, 8 threads, 16 tables, 100,000 rows per table
- run mode: 30s warmup + 300s measured run per workload, order `master(A1) -> branch(B1) -> branch(B2) -> master(A2)`
- workloads: `oltp_point_select`, `oltp_read_only`, `oltp_update_non_index`, `oltp_insert`
- all measured sysbench runs reported `ignored_errors=0`

| workload | master avg TPS/QPS | branch avg TPS/QPS | delta | master p95 | branch p95 | raw runs (master -> branch) |
| --- | ---: | ---: | ---: | ---: | ---: | --- |
| `oltp_point_select` | 125,025.57 | 127,788.74 | +2.21% | 0.11 ms | 0.11 ms | 127,028.12 / 123,023.03 -> 128,847.35 / 126,730.12 |
| `oltp_read_only` | 4,977.34 | 4,965.74 | -0.23% | 2.05 ms | 2.01 ms | 5,053.53 / 4,901.14 -> 5,046.79 / 4,884.68 |
| `oltp_update_non_index` | 2,838.90 | 2,833.89 | -0.18% | 2.97 ms | 2.97 ms | 2,842.87 / 2,834.94 -> 2,830.08 / 2,837.71 |
| `oltp_insert` | 2,264.27 | 2,353.93 | +3.96% | 9.41 ms | 8.74 ms | 2,185.74 / 2,342.81 -> 2,352.38 / 2,355.48 |

The cloud-host result shows no material performance regression. Read-only/update workloads are within noise level, point-select is slightly higher on the branch, and insert is higher in this run.

## Upgrade Compatibility
No storage format, protocol, or upgrade procedure impact.

The query-response-time diagnostic interface is removed
- `oceanbase.__all_virtual_query_response_time`
- `information_schema.QUERY_RESPONSE_TIME`
- `V$OB_QUERY_RESPONSE_TIME_HISTOGRAM`
- `SHOW QUERY_RESPONSE_TIME`

## Other Information
- Related work item
- MR


hnwyllmm:
The mapping Dima issue is about optimizing small memory blocks.

footka (approved):

PR #1049, opened 2026-07-14
Remove multi-log-stream architecture

footka:
## Release Note


footka:
The mapping Dima issue is about removing the multi-log stream code.

hnwyllmm (approved):

PR #1051, opened 2026-07-14
Remove ObStorageLoggerManager and move slog lifecycle into storage meta service

hnwyllmm:
## Task Description
Remove ObStorageLoggerManager and redistribute its responsibilities to the components that own them.

## Solution Description
- Move server slogger ownership and lifecycle management to ObServerStorageMetaService.
- Pass slog initialization settings directly to ObStorageLogger::init.
- Move disk reservation logic and SLOG_RESERVED_DISK_SIZE to ObServerStorageMetaService.
- Move slog item and buffer allocation/freeing into ObStorageLogger.
- Replace Manager-based tenant slog path access with Router-root-based path construction.
- Update checkpoint, meta-store, block-manager, and unit tests for the new ownership model.


hnwyllmm:
The Dima issue is about changing the standard log buffer to be allocated on-demand.

footka (approved):

PR #1088, opened 2026-07-16
Remove the schema sys cache and formalize schema fetch dependencies

hnwyllmm:
## Task Description
The schema cache maintained a dedicated sys-cache path for necessary system schemas. This introduced duplicated cache logic and implicit dependencies between schema fetching, core tables, and normal system tables.
Remove the sys cache while preserving bootstrap safety and schema visibility correctness. Schema-fetch dependencies also need to be explicitly defined so that bootstrap and on-demand schema loading do not introduce dependency cycles.

## Solution Description
- Removed the dedicated schema sys cache and unified normal schema caching through the standard schema cache.
- Retained a bootstrap cache as the fallback source before schemas can be loaded through the standard path.
- Formalized and reorganized schema-fetch dependency table IDs.
- Skipped partition, constraint, foreign-key, and trigger metadata queries for dependency tables to avoid cyclic schema access.
- Added normal_schema_version to __all_global_stat as the visibility watermark for normal schemas.
- Advanced the normal schema watermark after successful DDL transaction completion.
- Added bootstrap-safe handling for schema version and optimizer statistics queries.
- Updated INFORMATION_SCHEMA.TABLES and INFORMATION_SCHEMA.COLUMNS to include the core history tables.
- Updated affected mysqltest results for the inner-table definition changes.
Validated with a fresh bootstrap and the following mysqltest cases
- information_schema
- information_schema.information_schema_desc
- histogram.stats_farm
- inner_table.show_sys_tables_in_sys


hnwyllmm:
The mapping Dima issue is about moving schema-related core storage tables from no-swap cache to regular cache.

footka:
> **src/rootserver/virtual_table/ob_virtual_core_inner_table.cpp:93**

Use a local struct instead of having so many arrays.

footka:
> **src/rootserver/ob_load_inner_table_schema_executor.cpp:240**

Should the following sys tables be written as normal?

footka:
> **src/share/schema/ob_schema_cache.cpp:430**

What else is this mem_context_ used for?

footka:
> **src/share/ob_global_stat_proxy.cpp:167**

What's the difference between advance and set?

hnwyllmm:
> **src/rootserver/ob_load_inner_table_schema_executor.cpp:240**

This is called during bootstrap when reading the schema from inner_table_schema.

In the current code, `sys_schema_version` semantically includes `core_schema_version`, so `normal_schema_version` includes `sys_schema_version`.

If we don't add it here, it relies on being updated later by DDL statements. Adding it here doesn't make a big difference; it's mainly for semantic alignment.

hnwyllmm:
> **src/share/ob_global_stat_proxy.cpp:167**

Since we initially considered that regular DDL operations might enter a concurrent state, we added `advance` to only push forward to a higher version, and `set` to directly set the version.

Later, while handling the DDL wait issue, we found there is no scenario involving a rollback. The plan is to modify this by removing the `advance_normal_schema_version` version, keeping the `set` method, and adding a forward-only push mechanism.

hnwyllmm:
> **src/share/schema/ob_schema_cache.cpp:430**

Previously, I thought `bootstrap_cache_` was only used during startup and would be cleared later, so I didn't pay attention to it.

Now that the schema is fully independent, it's not needed during startup either, so I plan to remove it.

hnwyllmm:
> **src/rootserver/virtual_table/ob_virtual_core_inner_table.cpp:93**

Modified

footka:
> **src/rootserver/ob_load_inner_table_schema_executor.cpp:240**

Research on `set_sys_schema_version`

footka:
> **src/rootserver/virtual_table/ob_virtual_core_inner_table.cpp:100**

TO_STRING_EMPTY

hnwyllmm:
> **src/rootserver/ob_load_inner_table_schema_executor.cpp:240**

sys_schema_version is used elsewhere, so it should not be changed.

hnwyllmm:
> **src/share/ob_global_stat_proxy.cpp:167**

Already only keeping the set operation, consistent with the logic of `set_sys_schema_version` and `set_core_schema_version`.

hnwyllmm:
> **src/rootserver/virtual_table/ob_virtual_core_inner_table.cpp:100**

Replaced TO_STRING_EMPTY.

hnwyllmm:
> **src/share/schema/ob_schema_cache.cpp:430**

Cleared.

footka (approved):

PR #1097, opened 2026-07-17
Remove partition split && dynamic partition

footka:


footka:
The mapping Dima issue is about removing partition splitting and dynamic partitioning.

hnwyllmm (approved):

PR #1100, opened 2026-07-17
Advertise autocommit in MySQL handshake

LINxiansheng:
## Task Description
Set SERVER_STATUS_AUTOCOMMIT in the initial HandshakeV10 packet to match seekdb's default autocommit state. This prevents clients such as PyMySQL from incorrectly skipping SET AUTOCOMMIT=0.

## Solution Description
Add wire-format tests for the default status and explicit overrides.

## Passed Regressions
core test


LINxiansheng:
The mapping Dima issue is about fixing the autocommit handshake state.

footka:
> **deps/oblib/src/rpc/obmysql/obsm_struct.h:115**

Why not just use int?

footka:
> **src/observer/mysql/obsm_conn_callback.cpp:124**

Same as above, what's the purpose of converting to bool?

footka:
> **src/sql/session/ob_basic_session_info.cpp:4916**

It's recommended to implement this via a patch. Avoid modifying the `load_all_sys_vars` function internally. Instead, have the caller actively make the changes after calling `load`.

footka:
> **src/observer/mysql/obsm_conn_callback.cpp:124**

And the best implementation would still be to access it through the same schema version number, but currently it's being forced to change.

footka:
> **src/observer/mysql/obsm_conn_callback.cpp:124**

We can record a schema_version in smconnect, and both the handshake packet and subsequent authentication and loading of system variables will use this version for access.

LINxiansheng:
> **deps/oblib/src/rpc/obmysql/obsm_struct.h:115**

done

LINxiansheng:
> **src/sql/session/ob_basic_session_info.cpp:4916**

done

LINxiansheng:
> **src/observer/mysql/obsm_conn_callback.cpp:124**

done

footka (approved):

PR #1158, opened 2026-07-21
Remove obsolete SeekDB inner tables

footka:
## Release Note


footka:
The Dima mapping issue is about deletion using internal tables as a starting point.

hnwyllmm (approved):

PR #1165, opened 2026-07-22
refactor: remove dup table support

LINxiansheng:
## Task Description
refactor: remove dup table support

## Solution Description
remove dup table schema, DDL, parser and RPC definitions
remove optimizer, plan cache and replica selection logic
remove transaction, lease, retry, log and error handling
delete related tests, configurations and golden files
preserve standard ON DUPLICATE KEY UPDATE semantics

## Passed Regressions
core test

## Release Note


LINxiansheng:
The mapping Dima issue is [delete dup table](

footka (approved):

PR #1167, opened 2026-07-22
Fixed seekdb macOS compilation bug

ep-12221:
## Task Description
Fixed compilation errors on macOS related to `curl` and `long long` data types.

## Solution Description
On macOS, the build no longer uses `curl`. Additionally, unified the argument types for `std::min` calls to resolve type mismatch warnings/errors.

## Passed Regressions
core-test

## Release Note

### DIFF
--- a/src/share/ob_telemetry.cpp
+++ b/src/share/ob_telemetry.cpp
@@ -21,11 +21,7 @@
 #include "share/config/ob_server_config.h"
 #include "share/ob_encryption_util.h"
 #include "common/ob_version_def.h"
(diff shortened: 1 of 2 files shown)


ep-12221:
The mapping Dima issue is [Mac seekdb compilation failure due to missing curl]

footka:
> **src/share/ob_telemetry.cpp:277**

Why not address this directly? This is the reporting logic for installed device count, which needs to be supported.

footka (approved):

PR #1169, opened 2026-07-22
[seekdb][storage] Event-driven snapshot GC renewal

hnwyllmm:
## Task Description
ObMajorMergeInfoDetector periodically renews snapshot_gc_scn even when the tenant has not produced new historical row versions. This causes unnecessary internal transactions and writes in idle or low-load seekdb instances.

## Solution Description
- Keep the existing snapshot_gc_scn transaction and locking protocol, but start renewal only after the service actually becomes primary.
- Add a tenant-local CAS-max pending history SCN. Concurrent mini-merge and upper_trans_version events are coalesced without increasing the renewal frequency.
- Publish events after successful mini merge for UPDATE/DELETE history, with an exact filter for the __all_global_stat.snapshot_gc_scn self-update row.
- Publish an event when an SSTable upper_trans_version changes from INT64_MAX to a finite value; routine old-SSTable GC does not trigger it.
- While pending history exists, renew at a fixed 10-second interval. undo_retention only participates in the coverage check and does not affect scheduling.
- Preserve pending state across pause/resume in the same process and retry renewal failures at the same fixed interval.
- Remove alarms that assume snapshot_gc_scn must continuously advance.

## Passed Regressions
- Full debug build: ob-make
- Module-layer dependency check: no upward edges
- Unit target build: ob-make test_tenant_freeze_info_mgr
- Unit test: test_tenant_freeze_info_mgr, 2/2 passed
- Concurrency repeat: test_tenant_freeze_info_mgr --gtest_repeat=100 --gtest_break_on_failure, all passed
- git diff --check

## Upgrade Compatibility
Compatible. This change does not modify the SSTable format, system-table schema, or persistent metadata. Pending event state exists only in tenant memory and is rebuilt by primary catch-up after restart.

## Other Information
This MR contains one focused commit: e4dce8ab0b9 make snapshot gc renewal event-driven.


hnwyllmm:
The mapping Dima issue is about optimizing snapshot_gc_scn.

footka (approved):

PR #1188, opened 2026-07-27
Set ObEmptyReadBucket capacity at fixed 16K buckets

hnwyllmm:
## Task Description
Performance testing shows that the size of ObEmptyReadBucket has minimal impact on performance. Fix its capacity at 16,384 buckets, consuming approximately 640 KiB, instead of dynamically scaling the bucket count based on server memory.

## Release Note

### DIFF
--- a/src/storage/access/ob_empty_read_bucket.cpp
+++ b/src/storage/access/ob_empty_read_bucket.cpp
@@ -15,7 +15,6 @@
  */
 
 #include "ob_empty_read_bucket.h"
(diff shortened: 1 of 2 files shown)


hnwyllmm:
The Dima mapping issue is about optimizing the bucket capacity for the Bloom Filter trigger counter.

hnwyllmm:
After building and testing, we evaluated the performance impact (85: origin, 86: fixed 16K buckets).

image

hnwyllmm:
image

footka (approved):

PR #1192, opened 2026-07-27
Fix vsag-related mac case

ep-12221:
## Task Description
Fix issues related to the vsag-related mac case.

## Solution Description
Primarily addressed floating-point problems. Memory estimation logic was changed to verify no errors. Some issues required changes to vsag, so those cases are still not passing.

## Passed Regressions
core-test

## Release Note


ep-12221:
The mapping Dima issue is related to vector fixes for the seekdb mac mini-farm.

footka (approved):

PR #1193, opened 2026-07-27
Fix uniform/normal distribution function for macOS

ep-12221:
## Task Description
The STL distribution function behaves differently between macOS and Linux, leading to inconsistent results.

## Solution Description
Rewrote the distribution function to avoid using the STL implementation, ensuring consistent behavior across platforms.

## Passed Regressions
core-test

## Other Information
- Fixed branches: master
- Related links: [DIMA-2026072300117675908]()


ep-12221:
The mapping Dima issue is about random fixes related to the seekdb mac mini farm.

hnwyllmm:
> **src/sql/engine/expr/ob_deterministic_distribution.cpp:50**

return min_value + offset;

??? Is there any difference?

ep-12221:
> **src/sql/engine/expr/ob_deterministic_distribution.cpp:50**

No difference, already updated.

hnwyllmm (approved):

PR #1196, opened 2026-07-27
[seekdb][deps] Switch al8 x86 SDKs to source submodule

hnwyllmm:
## Task Description
Currently, the development library dependencies are recorded via al8 x86_64 RPMs under `deps/init`, leading to unclear maintenance boundaries between source code versions and package versions.

## Solution Description
Added a `deps/dependencies` submodule pinned to the uploaded and pushed source dependency versions. `build.sh` now builds the Linux x86_64 source SDK during initialization and uses the corresponding installation prefix. Removed development library package records from `oceanbase.al8.x86_64.deps [deps]`, while retaining tool dependencies. For VSAG linking items built from source, linking is now optional and based on actual build artifacts, specifically for x86_64.

## Passed Regressions
Passed `bash -n build.sh`, `git diff --check`, and `git diff --cached --check`. Verification of source dependency repository origin locking and Linux x86_64 dependency building is complete.

## Upgrade Compatibility
Only adjusts the method of obtaining development dependencies; does not change runtime protocols. This demo only covers al8 x86_64.

## Other Information
This MR contains only one commit relative to the master branch. Dependency description files for other platforms remain unchanged. Untracked local environment hint files are not included in the commit.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "deps/dependencies"]
+	path = deps/dependencies
+	url = ../seekdb-dependencies.git
--- a/build.sh
+++ b/build.sh
@@ -41,6 +41,20 @@ function echo_err() {
   echo -e "[build.sh][ERROR] $@" 1>&2
 }
 
+# Keep all native build tools and helper programs on the repository toolchain.
+# In particular, ICU and other source builds execute freshly-built host tools;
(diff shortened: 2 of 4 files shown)


hnwyllmm:
The issue with the mapping Dima is that it depends on development libraries in source code form.

PR #1200, opened 2026-07-27
Fix mac val cast problem

ep-12221:
## Task Description
Static cast long long overflow.

## Solution Description
Enhance range checking.

## Passed Regressions
core-test


ep-12221:
The mapping Dima issue is about the precision conversion problem in Mac seekdb's small farm.

hnwyllmm:
[P2] In `round_floating_to_uint64`, the checks for `rounded <= INT64_MIN_AS_DOUBLE` and `rounded >= UINT64_UPPER_BOUND_AS_DOUBLE` are handled before checking `is_column_convert`. In non-strict mode, `OB_DATA_OUT_OF_RANGE` is downgraded to a warning, but the current `out_val` is still persisted. This leads to incorrect results for extreme values in unsigned columns: a DOUBLE value of `-2e28` gets written as `2^63` (MySQL should clamp it to `0`), and a DOUBLE value of `+2e28` gets written as `INT64_MAX` (MySQL should clamp it to `UINT64_MAX`; the FLOAT path writes it as `2^63`). MySQL's `Field_longlong::store(double)` handles negative numbers and upper bounds according to unsigned column semantics first: https://github.com/mysql/mysql-server/blob/8.0/sql/field.cc. It is suggested to move the column-convert branch earlier, so negative numbers uniformly output 0 and values reaching 2^64 output UINT64_MAX. The non-column-convert branch should retain the existing differences between FLOAT/DOUBLE CAST AS UNSIGNED. Also, add test cases for FLOAT/DOUBLE `±2e28 -> BIGINT UNSIGNED` column conversion, as the current unit tests do not cover these two boundaries.

hnwyllmm (approved):

hnwyllmm (approved):

PR #1207, opened 2026-07-28
Remove Inner SQL Connection and Session Pool Caching

hnwyllmm:


hnwyllmm:
The mapping Dima issue is about removing ObInnerSQLConnectionPool and using it only as needed.

footka:
> **src/sql/hybrid_search/ob_hybrid_search_executor.cpp:142**

Why is reference counting still needed?

footka:
> **src/observer/ob_inner_sql_connection.cpp:1704**

What's special about this release, and why is it called release_inner_sql_connection_for_proxy? What's its relationship with the proxy?

hnwyllmm:
> **src/observer/ob_inner_sql_connection.cpp:1704**

This provides a Proxy for Oblib to resolve the WEAK_SYMBOL function dependencies from upper layers. There's no special cleanup logic, just an unref, used to isolate the dependencies between oblib and observer.

hnwyllmm:
> **src/sql/hybrid_search/ob_hybrid_search_executor.cpp:142**

Reference counting is unrelated to the connection pool; it ensures that the Conn is not released prematurely while a ReadResult is using the Session inside that Conn. This is the same function originally used for release—only the call location has been moved.

hnwyllmm:
> **src/observer/ob_inner_sql_connection.cpp:1704**

The smart pointer release has been unified, and `release_inner_sql_connection_for_proxy` is no longer needed and has been removed.

hnwyllmm:
> **src/sql/hybrid_search/ob_hybrid_search_executor.cpp:142**

Changed to use ObSharedGuard<T> smart pointer for allocation and deallocation.

footka:
> **deps/oblib/src/common/mysqlclient/ob_single_connection_proxy.h:66**

Using a guard in a local function scope is relatively safe, but using it as a class member can be dangerous.

footka (approved):

PR #1209, opened 2026-07-28
fix: handle stale routine definers in macOS mysqltests

ep-12221:
## Task Description
Fix issues including unstable sorting and non-existent commands on macOS.

## Passed Regressions
core-test


ep-12221:
The Dima mapping issue is related to SSH fixes for the seekdb mac small farm.

footka (approved):

PR #1210, opened 2026-07-28
Trim unsupported seekdb features (driven by dead config)

footka:


footka:
The mapping Dima issue is about configuration-driven deletion.

hnwyllmm (approved):

PR #1223, opened 2026-07-30
Remove JIT code

ep-12221:


ep-12221:
The mapping Dima issue is about removing the Mac-related code for PL/JIT.

PR #1225, opened 2026-07-30
[seekdb][change stream] Fix refresh consistency

hnwyllmm:
## Task Description
Change Stream previously used a single refresh SCN for both Worker-applied progress and user-visible refresh completion. This could lead to issues where the Fetcher periodically advanced this value without a Worker commit, while the Dispatcher also used it to skip transactions during recovery. Concurrent asynchronous index creation or deletion could therefore cause a refresh to return before the corresponding schema generation and log work were safely completed, or allow log reclamation to advance using a mismatched schema decision.

## Solution Description
- Split the persisted Worker-applied waterline from the in-memory refresh-visible waterline. The Dispatcher now skips transactions only using the persisted applied SCN.
- Added a Fetcher-owned, versioned asynchronous index schema state, including `ready_schema_version` and `last_no_async_index_drained_schema_version`.
- Modified `min_dep_lsn` advancement to require an exact match between the runtime schema version and the Fetcher's ready state.
- Capture refresh targets with a single deadline using schema version, GTS, and an LSN fence. Asynchronous refresh is only completed after the fence is processed and the Dispatcher barrier is committed, or after a later generation with no asynchronous operations is fully drained.
- Reset refresh state from the persisted applied progress during startup/recovery, and prevent refresh publication while the recovery baseline is incomplete.
- Removed the Fetcher's periodic GTS-based refresh advancement. The refresh virtual table now reads the visible waterline from `ChangeStreamMgr`.

## Passed Regressions
- ob-make
- vector_async.async_vec_cs_refresh_consistency
- vector_async.async_vec_manual_refresh
- vector_async.async_vec_ddl_drop_rebuild
- vector_async.ddl_delete_rebuild_index
- vector_async.async_vec_cs_fetcher_active_idle

## Upgrade Compatibility
No persistent metadata format changes. The existing `change_stream_refresh_scn` global-state item is retained as the Worker-applied waterline. The user-visible refresh waterline is maintained in memory and is initialized from this persisted value during startup or recovery.

## Other Information
This MR contains a single commit: `bdd5e6f9865 fix change stream refresh consistency`.

## Release Note
Fixed an issue in Change Stream where concurrent asynchronous index operations could cause a refresh to complete before the associated schema and log work was finished, or lead to log reclamation using an incorrect schema state, ensuring refresh operations are now consistent with the actual completion of background work.


hnwyllmm:
The Dima issue is about optimizing change stream refresh.

PR #1234, opened 2026-07-31
Fix mini mode memory and thread regressions

footka:


footka:
The mapping Dima issue is about memory overhead regression.

hnwyllmm (approved):

PR #1243, opened 2026-08-03
Fix GIS and sort problem in macOS

ep-12221:


ep-12221:
The mapping Dima issue is related to GIS problems on the Mac seekdb small farm.

footka (approved):

PR #1245, opened 2026-08-03
Change GIS to plugin

LINxiansheng:
## Task Description
Change GIS to plugin

## Solution Description
Move GIS functionality out of the seekdb core and into the standalone seekdb_gis.so plugin. Add a stable plugin SPI, core compatibility shims, plugin validation, and build tests so seekdb can run without built-in GIS dependencies

## Passed Regressions
core test


LINxiansheng:
The mapping Dima issue is [delete GIS](

PR #1246, opened 2026-08-03
Migrate Mysql protocol and NIO to Rust

hnwyllmm:
## Task Description
Migrate mysql protocol and nio networking part to Rust.

## Solution Description
A new folder named rust is created under the codebase of seekdb, where a crate sql-nio can be found. This crate implements mysql protocol for encoding and decoding mysql packets and nio part with mio. In the original version of seekdb, linux uses SO_REUSEPORT for scaling incoming connections whereas mac/windows delivers all connections to only one thread. In this migration, all platform uses one thread to accept incoming connections and dispatch them to multiple io threads.

## Passed Regressions
This version shows same performance on sysbench compared to baseline.

## Release Note


hnwyllmm:
The mapping Dima issue is [Migrate seekdb to Rust](

footka (approved):

PR #1249, opened 2026-08-04
Remove virtual thread case

ep-12221:


ep-12221:
The mapping Dima issue is about removing the virtual_thread case.

footka (approved):

PR #1250, opened 2026-08-04
Fix GIS case

ep-12221:


ep-12221:
The mapping Dima issue is about adjusting the GIS case.

footka (approved):

PR #1252, opened 2026-08-04
[seekdb][share] Avoid idle internal table polling

hnwyllmm:
## Task Description
Freeze, Timezone, DBMS Job, and DBMS Scheduler periodically query internal tables even when the tables have not changed. These idle reads continuously consume CPU and cache resources.

## Solution Description
Maintain a process-local change sequence for each registered internal table. Transaction commit increments the corresponding sequence, while consumers keep their existing timers and skip internal SQL when the captured sequence is unchanged.
Seal the table registry after all modules finish initialization and before consumer timers start. Keep fail-open refresh semantics when sequence lookup fails. DBMS Job and DBMS Scheduler use full reconciliation on table changes and preserve retry/self-healing behavior for transient scheduling failures.

## Passed Regressions
- `git diff --check`
- `ob-make seekdb`
- `ob-make test_internal_table_change_notifier`
- `test_internal_table_change_notifier`: 4/4 passed
- seekdb startup smoke: notifier sealed with 9 tables and server startup reached 14/14

## Upgrade Compatibility
Compatible. The change sequence is process-local and introduces no persistent data-format or upgrade change.

## Other Information
- This MR contains one commit: `e906c59d6f0 optimize: avoid idle inner table polling`.
- SQLite-backed `refresh_merge_info` remains periodically refreshed and is not gated by the internal-table sequence.
- No background thread pool experiment is included.
- Related internal link: DIMA-2026080300117949904


hnwyllmm:
The mapping Dima issue is related to inner SQL optimization.

PR #1256, opened 2026-08-05
refactor(sql): remove non-vectorized execution mode

LINxiansheng:
## Task Description
Remove the non-vectorized execution mode from the SQL engine.

## Solution Description
1. Remove the `rowsets_enabled` configuration, hint, and plan cache state.
2. Always use the vectorized cost model and execution path.
3. Delete normal cost parameters and legacy scalar-mode tests.
4. Retain single-row boundary coverage by using a vectorized batch size of 1.

## Passed Regressions
core test

## Release Note


LINxiansheng:
The mapping Dima issue is [deleting non-vectorized code].

footka (approved):

PR #1262, opened 2026-08-07
[seekdb][memory] Refine jemalloc integration and module memory limits

hnwyllmm:
## Task Description
This change refines the integration of the packaged jemalloc library and decouples the memory limits for key modules (KVCACHE, Memstore, Vector) from the global `memory_budget` percentage parameters. It introduces absolute limit parameters for these modules, allowing for independent and more intuitive configuration of their capacities.

## Solution Description
The implementation involves several key changes:
1.  **CMake Integration:** The packaged jemalloc setup, previously maintained in a separate CMake file, has been inlined and simplified at the root CMake level.
2.  **API Clarification:** The cache-node accounting API has been renamed to clarify that it only releases memory accounting, avoiding confusion with object destruction.
3.  **New Absolute Parameters:** Absolute configuration parameters (`kvcache_memory_limit`, `memstore_memory_limit`, `vector_memory_limit`) have been introduced.
4.  **Automatic Sizing:** When an absolute limit is set to 0, it is automatically derived from the cgroup-aware effective system memory (KVCACHE: 25%, Memstore: 20%, Vector: 10%).
5.  **Module Migration:** The KVCACHE washing/prewarming logic, Memstore freezing/throttling, Vector allocation checks, and related mysqltest configurations have been migrated to use the new absolute limits.
6.  **Parameter Cleanup:** The retired percentage-based parameters for Memstore and Vector have been removed.

Core changes to module limits relative to the new base `_memory_budget` (`B`) include:
-   **KV Cache:** Decoupled from `B`. Default is `30% × P` (physical memory), configurable via `kvcache_memory_limit`.
-   **Memstore & Vector:** Default raised to `80% × B`, configurable via absolute `memstore_memory_limit` and `vector_memory_limit`. Total consumption is constrained by the TxShare limit.
-   **TxData & MDS:** Module and auto-freeze thresholds doubled (to `40%/10% × B` and `20%/4% × B` respectively) to better adapt to smaller `B` values.
-   **TxShare:** Limit formula adjusted (`B<=4G` ? `110%B` : `130%B`).
-   **Log Allocator:** CLog pool uses `30% × B`, Replay uses `5% × B`.
-   **Other Modules (Plan Cache, SQL Work Area, DTL, Compaction, etc.):** Calculation formulas remain the same but now use `B` as the base instead of the old `memory_limit` (`L`).

## Passed Regressions
- test_utility: 12/12 passed.

## Upgrade Compatibility
Configuration names and semantics have changed. Old percentage-based settings for Memstore/Vector and old limit names for KV/Vector must be migrated to the new absolute parameters: `kvcache_memory_limit`, `memstore_memory_limit`, and `vector_memory_limit`. A non-zero value for these parameters is interpreted as an absolute capacity in bytes. Setting them to 0 enables the new automatic sizing behavior based on system memory.

## Other Information
This MR contains one commit and does not include local untracked test artifacts.

## Release Note
Refined jemalloc integration and memory management. Introduced absolute configuration parameters (`kvcache_memory_limit`, `memstore_memory_limit`, `vector_memory_limit`) to independently control module memory capacities, replacing previous percentage-based coupling to `memory_budget`. When set to 0, limits are automatically sized based on available system memory. Removed deprecated percentage parameters for Memstore and Vector. Requires configuration migration for existing deployments using the old parameters.


hnwyllmm:
The mapping Dima issue is "Adjustments to configuration items and partial code optimization after jemalloc adaptation."

footka (approved):

LINxiansheng (approved):

Issue #1267, opened 2026-08-08
Retry auto-increment sequence allocation internally on OB_EAGAIN instead of returning error 4023 to client


footka:
## Description

During the TransactionSize prepare phase of the `bigtrans` workload, under high-concurrency table creation and batch writes, a `BatchUpdateException: Try again` (SeekDB error code 4023) is occasionally returned via JDBC, causing the entire workload to fail.

## Root Cause

When `ObTabletAutoincMgr::fetch_new_range` calls the local `ObTabletAutoincSeqService` to allocate a new sequence range, the underlying log submission may transiently return `OB_EAGAIN`. The original implementation directly propagated this retryable error to the SQL/JDBC layer, failing to absorb transient contention on the server side.

## Expected Behavior

For errors identified as retryable, the system should retry within the current worker's timeout and cancellation constraints, using the existing `RETRY_INTERVAL`. Timeouts, request cancellations, or non-retryable errors should still be returned as-is. Retries must not be infinite, and the error must not be bypassed on the test side.

## Proposed Solution

Add a retry loop within `fetch_new_range`, constrained by the worker's deadline and status checks. On failure, reset the interval, wait for `RETRY_INTERVAL`, and re-attempt the allocation. On success, reuse the existing logic for range validation and assignment.

## Verification

1.  In a real SeekDB environment, inject a single `OB_EAGAIN` error: the client SQL did not fail, and subsequent allocation succeeded.
2.  Run the official `bigtrans` full prepare + 300-second workload using the patched binary. All 16 cases completed, `run.sh` returned 0 and output `bigtrans_lite PASS`.
3.  The full regression suite did not naturally trigger the auto-increment 4023 error, so this PASS result is not considered standalone evidence for retry branch coverage.

PR #1285, opened 2026-08-12
fix(storage): retry transient tablet autoincrement failures

footka:
## Task Description
This change addresses a regression introduced by a previous commit that removed retry logic for transient failures during tablet autoincrement operations.

## Solution Description
The regression was introduced by commit b1013d2d ("Trim unsupported seekdb features"), which replaced the RPC sequence-cache fetch with a local service and inadvertently dropped the existing retry loop. This fix restores the retry mechanism for transient failures in the tablet autoincrement process.

## Release Note
Fix: Restore retry logic for transient failures during tablet autoincrement operations, addressing a regression introduced by a previous code change.


footka:
The mapping Dima issue is [Tablet auto-increment sequence encounters OB_EAGAIN and returns 4023 to the client](https://github.com/oceanbase/seekdb/issues/1267).

hnwyllmm (approved):

PR #1290, opened 2026-08-12
Restore runtime gate for standby gRPC service

footka:


footka:
The mapping Dima issue is [Restore the standby gRPC service runtime switch and disable the listener by default](https://project.alipay.com/workItem?workItemId=2026081100118157754).

hnwyllmm (approved):

PR #1293, opened 2026-08-12
Deduplicate partition columns for local index build

footka:


footka:
The corresponding Dima issue is [[seekdb] Duplicate local index backfill columns when generated columns and their dependent columns are used together as partition keys](https://project.alipay.com/workItem?workItemId=2026081200118194652).

Issue #1270, opened 2026-08-09
Restore and Refactor SeekDB Standby Capability


footka:
## Description
Restore the complete standby capability that was temporarily removed in commit `9c6764cc8704a8a022fea2d878458bdc9f7bd8e0`, and perform a modular refactoring based on the current master branch.

## Scope
1.  Restore gRPC-based log synchronization, SSTable baseline/bootstrap, incremental log replay, cascading standbys, multiple standbys, dynamic source, and TLS.
2.  Retain the standby's read-only service, supporting one-way in-process standby promotion. Primary-to-standby switchover uses write fencing and becomes effective after restarting with the persisted pending role.
3.  Converge role state, switchover control, synchronization, bootstrap, gRPC service, and background tasks into `src/standby`. The primary path should depend on boot profile/capability and a unified write gate.
4.  Standby performs autonomous freeze and merge; do not restore primary-standby checksum and the binding of freeze/merge operations.
5.  Maintain the existing test assertions in `t/standby` for position, read-only status, data consistency, and role; do not bypass issues by relaxing test cases.

footka:
The code has been migrated to the local branch task/2026081000118111905, based on origin/master f84169fbd80.

Cherry-picked commits:
- 5789a6ca185 Restore and encapsulate standby support
- 94ce3308867 refactor standby replication lifecycle
- 8dc58941d2d Refactor standby role transitions for restart activation
- 0560916c033 Implement one-way online standby promotion
- a2a30a72ff8 Deepen standby module ownership

Verification: Module boundary checks pass; the entire tree is consistent with the temporary vertex after rebase. Remote Bazel compilation is blocked by a missing private header declaration for `ob_json_base.h`, introduced by commit 289ae19d97b in origin/master, which is unrelated to the standby changes.

PR #1294, opened 2026-08-12
Recovery Physical Standby

footka:
## Task Description
1. Restore standby database capability.
2. Reference the PostgreSQL physical standby model, adopting an asymmetric role switch model to simplify implementation complexity.
   - PRIMARY -> STANDBY: First, disable writes online and persist the target role; the change takes effect after restart.
   - STANDBY -> PRIMARY: In-process one-way promotion, no restart required, read-only service on the standby is not interrupted.
   - External orchestration is responsible for node selection, command execution, traffic switching, source configuration, stopping, and restarting.
   - The kernel is responsible for write isolation, switch boundaries, log catch-up, role persistence, and startup recovery.
   - In case of a switch failure, it is preferable to have no primary temporarily rather than allowing two writable primaries.

## Release Note
- [DIMA-2026081000118111905](https://project.alipay.com/workItem?workItemId=2026081000118111905)


footka:
The corresponding Dima issue is [Restore and Refactor SeekDB Standby Capability](https://github.com/oceanbase/seekdb/issues/1270).

hnwyllmm (approved):

Issue #1268, opened 2026-08-08
SIGSEGV due to null pointer in T1_ReplaySrv0 during transaction log replay


chyujing:
## Description
SeekDB crashed with a SIGSEGV during startup.
Core file: `/data/log/bak_core/core-T1_ReplaySrv0-1624523-1786086302`

## Crash Point
Function: `oceanbase::pl::ObPLComposite::set_is_null(this=0x0, is_null=true)`
Source file: `src/pl/ob_pl_user_type.h:439`

## Key Call Stack
```
ObMulSourceTxDataNotifier::notify
-> ObTxCtx::notify_data_source_
-> ObTxCtx::replay_commit_info
-> ObTxReplayExecutor::replay_commit_info_
-> ObLSTxService::replay
-> ObLogReplayService::do_replay_task_
```

## Context
- Thread: T1_ReplaySrv0 (LWP 1624687)
- Transaction ID (tx_id): 1000866
- Notify type: TX_END
- `for_replay=true`
- `is_incomplete_replay=true`
- Binary: `seekdb-master-928324ce-20260807114446 build_debug`
- Machine link: http://:6525/data/1/obtool/seekdb/runs/seekdb-master-928324ce-20260807114446/instance-native/

PR #1299, opened 2026-08-13
[seekdb][MDS] Fix buffer context deep copy traversal

hnwyllmm:
## Task Description
Transaction log replay could encounter a null-pointer SIGSEGV in T1_ReplaySrv0. The issue was that `MdsFactory::deep_copy_buffer_ctx` compared the recursive tuple index directly with the sparse binding type ID. For MdsCtx, the binding type ID is 0 while its tuple index is 1, causing the function to incorrectly return success without allocating a copied context.

## Solution Description
Modified the traversal logic to iterate through every registered BufferCtx tuple entry and match the source context using the `BufferCtxBindingTypeId`. Added a success postcondition that rejects a null copied context. A regression test was also added to cover the MdsCtx binding-ID/tuple-index mismatch scenario.

## Passed Regressions
- ob-make test_tablet_create_mds_ctx
- test_tablet_create_mds_ctx: 2/2 cases passed
- The new regression case reproduced the `copied_ctx == nullptr` condition before the fix and passed after the fix.

## Upgrade Compatibility
No serialization format or public API changes. This fix only corrects the runtime BufferCtx type lookup logic and adds a defensive error check.

## Other Information
- This MR contains only commit 3c338fcacea.
- Core-test G4890001028 is running.
- Local untracked files are not included.
- Related internal link: DIMA-2026080700118077486

### DIFF
--- a/unittest/storage/test_tablet_create_mds_ctx.cpp
+++ b/unittest/storage/test_tablet_create_mds_ctx.cpp
@@ -21,7 +21,7 @@
 
 #include "lib/allocator/page_arena.h"
 #include "lib/ob_errno.h"
-#include "share/rc/ob_module_provider.h"
+#include "share/rc/ob_server_runtime.h"
 #include "storage/multi_data_source/compile_utility/compile_mapper.h"
 #include "storage/multi_data_source/mds_ctx.h"
 #include "storage/multi_data_source/ob_tablet_create_mds_ctx.h"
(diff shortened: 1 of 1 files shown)


hnwyllmm:
The mapping Dima issue is [core: SIGSEGV due to null pointer in T1_ReplaySrv0 during transaction log replay](https://github.com/oceanbase/seekdb/issues/1268).

PR #1301, opened 2026-08-14
[seekdb][PX] Downgrade admission to available workers

hnwyllmm:
## Task Description
DBMS_STATS internal SQL can request a Degree of Parallelism (DOP) higher than `parallel_servers_target`. The existing PX admission path would wait unless the complete request could be satisfied or no PX query was running. Under sustained concurrent PX traffic, smaller requests could repeatedly consume released workers and delay the large request until query timeout.

## Solution Description
Determine the admitted worker count once before execution from the currently available PX quota. Pass the plan's minimum worker count into the target monitor, admit `min(requested, available)` workers when that count satisfies the minimum, and otherwise keep waiting. Existing DFO assignment proportionally downgrades the query to the admitted count, and the worker count remains fixed for that execution.

## Passed Regressions
- `./build.sh release --init`
- `ob-make ob_sql`
- `git diff --check`

## Upgrade Compatibility
No schema, configuration, or external API compatibility impact. The change only adjusts the internal PX admission decision.

## Other Information
This MR contains one commit: 076ad53ea7d. The local untracked file named `2` is unrelated and is not included.

### DIFF
--- a/src/observer/ob_server.cpp
+++ b/src/observer/ob_server.cpp
@@ -43,7 +43,6 @@ int ObServer::get_lower_bound_freeze_info(const int64_t snapshot_version, share:
 #include "lib/ob_running_mode.h"
 #include "lib/task/ob_timer_monitor.h"
 #include "lib/task/ob_timer_service.h" // ObTimerService
-#include "lib/utility/utility.h"
 #include "observer/ob_server_utils.h"
 #include "observer/ob_server_options.h"
 #include "share/ob_timezone_mgr.h"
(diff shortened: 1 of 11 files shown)


hnwyllmm:
The mapping Dima issue is: [[SeekDB][ddltest_duomo_lite] DBMS_STATS.GATHER_SCHEMA_STATS triggers a 2013 disconnection, with the observer repeatedly experiencing log submission/freeze/compaction exceptions: -6268/-4023/-4012

PR #1302, opened 2026-08-14
[seekdb][optimizer] Retry schema stats after tablet replacement

hnwyllmm:
## Task Description
Concurrent DDL such as TRUNCATE can replace a table's tablet after GATHER_SCHEMA_STATS has obtained its schema. The block-count estimation then accesses the stale tablet id and returns OB_TABLET_NOT_EXIST(-4725), which is propagated to the caller and causes ddltest_duomo_lite strict failure.

## Solution Description
When prepare_gather_stats returns OB_TABLET_NOT_EXIST before statistics are written, report that the runtime schema needs refreshing. GATHER_SCHEMA_STATS refreshes the schema guard, rebuilds the current table's statistics parameters, and retries the table once. After refresh, the existing OB_TABLE_NOT_EXIST handling continues to skip a table that was actually dropped. The retry is limited to one attempt to avoid an unbounded loop.
A debug-sync point and mysqltest reproduce the race deterministically by truncating a table immediately before block-count estimation.

## Passed Regressions
- Full Bazel build passed.
- ddl.gather_schema_stats_truncate_retry deterministic mysqltest passed.
- ddltest_duomo_lite vector and hybrid phases ran for 600 seconds each and both strict summaries passed. All 8 GATHER_SCHEMA_STATS calls succeeded, with no OB_TABLET_NOT_EXIST(-4725).

## Upgrade Compatibility
No on-disk format, system-table schema, SQL syntax, or protocol change.

## Other Information
This MR contains one commit. The unrelated untracked file named 2 is excluded.


hnwyllmm:
The mapping Dima issue is: Under concurrent DDL/vector workload in DBMS_STATS, GATHER_SCHEMA_STATS returns OB_TABLET_NOT_EXIST(-4725), causing a strict failure in ddltest_duomo_lite.

PR #1303, opened 2026-08-14
fix: downgrade expected stability logs to warn

footka:
## Task Description
Downgrade expected stability logs from a higher severity level (likely `ERROR` or `INFO`) to `WARN`.

## Solution Description
Modified the log level for specific, expected stability-related log messages from a higher severity to `WARN` to reduce log noise and better reflect their non-critical nature.

## Release Note
Fixed an issue where expected stability events were logged at an inappropriately high severity level (e.g., ERROR/INFO). These logs are now correctly emitted at the WARN level.


footka:
The mapping Dima issue is [seekdb: Restore the expected path WARN log level introduced by cb3ca34ac637].

PR #1304, opened 2026-08-14
[standby] Use process UUID for promotion cycle detection

footka:


footka:
The mapping Dima issue is [[seekdb][standby] Fix false positives in primary-standby switch loop detection]

PR #1305, opened 2026-08-14
Validate SQLite guard before transaction cleanup

footka:
## Task Description
Validate the SQLite guard mechanism before performing transaction cleanup to prevent potential issues during the cleanup process.


footka:
The Dima mapping issue is that there is a missing validity check after a SQLite connection acquisition failure.

PR #1309, opened 2026-08-18
[seekdb][memory] respect cgroup memory limits in defaults

hnwyllmm:
## Task Description
Memory defaults were calculated from physical memory only. In containers with a smaller cgroup memory limit, this could produce oversized memory_budget and cache limits.

## Solution Description
- Detect finite cgroup v1/v2 memory limits, including mount-root mapping, escaped paths, ancestor limits, and unlimited sentinels.
- Use min(physical memory, cgroup memory limit) when deriving automatic memory_budget and KV-cache defaults.
- Expose memory_budget as a regular configuration item and update repository deployment, test, packaging, and documentation references.
- Upgrade jemalloc RPM dependencies on AL8, EL7, EL8, and EL9 for both aarch64 and x86_64 to build 32026080717. Windows and macOS remain unchanged.

## Passed Regressions
- git diff --check
- ./bazel.py build //src/oblib:_oblib_foundation_impl
- ./bazel.py build //unittest/oblib:_oblib_tests_unity_9
- ./bazel.py build //unittest/share:_share_tests_unity_1
- clang -fsyntax-only on generated src/share/config_config.unity.cpp
- isolated cgroup v1/v2 limit-detection smoke test

## Upgrade Compatibility
The hidden configuration name _memory_budget is replaced by the regular name memory_budget. Repository-owned configurations have been migrated; external configurations using _memory_budget must switch to memory_budget.

## Other Information
- The full //src/share:config build remains blocked by a pre-existing cache module header dependency issue; the modified production config Unity source passes standalone syntax validation.
- The untracked local file 2 is not included in this MR.

### DIFF
--- a/docs/developer-guide/en/android.md
+++ b/docs/developer-guide/en/android.md
@@ -81,9 +81,9 @@ adb shell "/data/local/tmp/seekdb --nodaemon \
 ```
 
 Explicit resource values are recommended on Android. When `memory_budget` is
-zero or omitted, it defaults to the greater of 1 GiB and 50% of cgroup or
-physical memory; size it together with the data and log files for the available
-device resources.
+zero or omitted, it targets 80% of cgroup or physical memory while reserving at
+least 1 GiB for the system when possible; its minimum automatic value is 1 GiB.
+Size it together with the data and log files for the available device resources.
 
 ### Forward ports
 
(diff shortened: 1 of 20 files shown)


hnwyllmm:
The Dima mapping issue is about [adding memory budget statistics for container memory limits].

LINxiansheng (approved):

footka (approved):

PR #1310, opened 2026-08-18
[seekdb][transaction] Allocate log callbacks on demand

hnwyllmm:
## Task Description
Transactions have three reserved redo log callbacks. After they are exhausted, the current implementation obtains extra callback groups from an LS-level object pool. The pool starts empty and is adjusted once per minute, so load bursts can fail with OB_TX_NOLOGCB(-6268) before the pool has expanded, delaying transaction log progress.

## Solution Description
- Allocate one ObTxLogCbGroup from the process allocator on demand when the transaction callback list is empty, while preserving _trx_max_log_cb_limit.
- Mark dynamically allocated groups explicitly, skip object-pool statistics and address inference, and release them when the transaction context resets.
- Stop periodic callback-pool sizing because the extra callback path no longer uses that pool.
- Base pending-log overflow detection on the per-transaction callback limit instead of pool occupancy.

## Passed Regressions
- git diff --check
- ob-make ob_storage
- CARGO_NET_OFFLINE=true ob-make seekdb

## Upgrade Compatibility
No storage format, schema, protocol, or configuration compatibility impact.

## Other Information
This MR contains one commit: fec8f5fc38e. No unrelated changes are included. Dynamic allocation follows the existing log_cb_lock_ critical section to preserve callback-list concurrency semantics.

### DIFF
--- a/src/storage/tx/ob_tx_ctx_log_cb_helper.cpp
+++ b/src/storage/tx/ob_tx_ctx_log_cb_helper.cpp
@@ -101,18 +101,16 @@ int ObTxCtx::get_log_cb_(ObTxLogCb *&log_cb)
       ret = OB_TX_NOLOGCB;
     } else if (OB_FAIL(alloc_log_cb(this, new_log_cb))) {
     } else {
-      {
-        ObSpinLockGuard guard(log_cb_lock_);
+      ObSpinLockGuard guard(log_cb_lock_);
+      allocated_log_cb_count = allocated_log_cb_count_;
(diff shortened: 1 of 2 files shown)


hnwyllmm:
The mapping Dima issue is "Under concurrent load, login reads __all_global_stat and triggers OB_ERR_SHARED_LOCK_CONFLICT (-6004), JDBC connection establishment returns 4012 Statement is timeout" for [SeekDB 1.4.0][batch_dml].

footka (approved):

PR #1312, opened 2026-08-18
fix(observer): return startup errors without invalid reload

footka:
## Task Description
This fix ensures that startup errors are returned correctly without triggering an invalid reload process.

## Solution Description
The change modifies the observer's startup error handling logic to bypass an invalid reload step when an error occurs, allowing the error to be surfaced directly.

## Release Note
Fixed an issue where observer startup errors were not properly returned due to an invalid reload step.


footka:
The mapping Dima issue is that a SIGSEGV is triggered when seekdb fails to validate startup resource parameters.

hnwyllmm (approved):

PR #1316, opened 2026-08-19
[seekdb][transaction] Retry GTS acquisition when not ready

hnwyllmm:
## Task Description
When the local GTS service has not prepared the next timestamp range, `ObTsMgr::get_gts` returns `OB_EAGAIN`. The snapshot acquisition path did not distinguish this temporary GTS state from generic `OB_EAGAIN`, so the error could escape the transaction layer and become a user-visible JDBC `Try again` during write pressure.
The original bigtrans_lite incident also contains PALF-side `OB_EAGAIN`. This change is intentionally limited to the GTS snapshot path and does not claim to retry or mask every PALF `OB_EAGAIN` source.

## Solution Description
- Restore `OB_GTS_NOT_READY` when nonblocking GTS acquisition returns `OB_EAGAIN`.
- Retry GTS acquisition internally with a 10 ms interval, for at most 1 second and no later than the request or transaction deadline.
- Release the transaction descriptor lock while synchronously waiting, then recheck interruption, abort, rollback and operation-sequence state after reacquiring it.
- Use the retrying snapshot path for transaction read snapshots, write-store-context snapshots and snapshot-version acquisition.
- Treat ID-not-ready errors as retryable in the existing MDS registration retry loop.

## Passed Regressions
- `git diff --check`
- `CARGO_NET_OFFLINE=true ob-make seekdb` (release build, compile and link passed)
The original CI endpoint had already been cleaned up, so the full bigtrans_lite workload was not replayed in this change.

## Upgrade Compatibility
No storage format, schema, wire protocol or configuration change. This only changes transaction-layer error classification and bounded retry behavior.

## Other Information
- MR contains one commit: `74e0e0b966e`.
- GTS polling interval is 10 ms; the bounded wait does not introduce an unbounded SQL retry.
- Generic PALF `OB_EAGAIN` handling remains outside this MR.
- Related links: DIMA-2026081800118330224

### DIFF
--- a/src/storage/tx/ob_trans_service.cpp
+++ b/src/storage/tx/ob_trans_service.cpp
@@ -399,13 +399,14 @@ int ObTransService::register_mds_into_tx(ObTxDesc &tx_desc,
   if (OB_SUCC(ret)) {
     do {
       ret = register_mds_into_ctx_(tx_desc, type, buf, buf_len, seq_no, register_flag);
-      if (OB_EAGAIN == ret && ObTimeUtil::current_time() >= tx_desc.expire_ts_) {
+      if ((OB_EAGAIN == ret || is_id_not_ready_err(ret))
+          && ObTimeUtil::current_time() >= tx_desc.expire_ts_) {
         ret = OB_TIMEOUT;
         TRANS_LOG(WARN, "register tx data timeout", KR(ret), K(tx_desc), K(type));
-      } else if (OB_EAGAIN == ret) {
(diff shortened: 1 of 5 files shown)


hnwyllmm:
The mapping Dima issue is that during bulk writes to wide tables, PALF/transaction logs continuously return error -4023, causing JDBC to receive a "Try again" message and resulting in UpdateIgnore initialization failure.

footka (approved):

PR #1318, opened 2026-08-20
fix(sql): pass has_lob_header to get_nullsafe_cmp_func for LOB comparisons

footka:


footka:
The mapping Dima issue is [[sqlancer] Misalignment of filter comparison parameters for LONGTEXT column pushdown: get_nullsafe_cmp_func has_lob_header always false causing mismatch (introduced in 2bf907c560f)]

hnwyllmm (approved):

PR #1319, opened 2026-08-20
[seekdb][sanity] Fix coredump in plan/PS cache and skip-index output

hnwyllmm:
## Task Description
Sanity Lite regression identified coredumps related to the seekdb source code: Plan Cache/PS Cache had object lifecycle issues during concurrent eviction and memory statistics, and the skip-index column output had an error in using the local buffer length.

## Solution Description
1. Complete cache object accounting and refresh within the Plan Cache node write lock scope to avoid accessing the object after releasing the last node reference.
2. PS Cache statistics callback now uses a snapshot of the item/info size to avoid dereferencing a destroyed ps item after the schema expires.
3. The skip-index column output delimiter now uses the actual extra print buffer capacity to avoid false positives in Sanity boundary checks and prevent abort.
4. Added regression coverage for Plan Cache memory accounting lifecycle.

## Passed Regressions
`git diff --check` passed. The fixed version has passed the previous Sanity Lite long-term regression, and the aforementioned coredumps have not reappeared. The local Bazel compilation of the current branch is blocked by an existing baseline issue with the repository's source/header inventory; this unrelated issue has not been modified.

## Upgrade Compatibility
No changes to persistent formats, protocols, or upgrade compatibility are involved.

## Other Information
This MR contains two commits: Plan/PS Cache lifecycle fixes and skip-index buffer length fixes. The Sanity PageArena detector issue itself is not included in this MR.

### DIFF
--- a/unittest/sql/plan_cache/test_memory_limit_accounting.cpp
+++ b/unittest/sql/plan_cache/test_memory_limit_accounting.cpp
@@ -22,7 +22,6 @@
 #include "sql/plan_cache/ob_i_lib_cache_object.h"
 #include "sql/plan_cache/ob_plan_cache.h"
 #include "sql/plan_cache/ob_prepare_stmt_struct.h"
-#include "sql/plan_cache/ob_ps_cache_callback.h"
 
 #define private public
 #include "sql/plan_cache/ob_ps_cache.h"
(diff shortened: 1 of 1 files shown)


hnwyllmm:
The mapping Dima issue is: [SeekDB] [Sanity] Lite regression found a coredump caused by the seekdb source code.

footka (approved):

PR #1320, opened 2026-08-20
[seekdb][Vector Index] Fix SINDI snapshot deserialization failure

hnwyllmm:
## Task Description
1. After creating a local SINDI index on a non-empty table, the first ANN query fails to deserialize the persisted snapshot and returns error 7604. OceanBase uses a callback sharding mechanism to implement `ObIStreamBuf`, but the VSAG `BufferStreamReader` incorrectly treats the current shard length as the full stream length, causing subsequent reads to go out of bounds and report a READ_ERROR.
2. Prevent duplicate final commit log submission. A transaction commit retry could arrive before the previous final state-log callback changed the downstream state, leading to the submission of a second `TX_COMMIT_LOG`. The duplicate callback later attempted to write already read-only transaction data, resulting in `OB_ERR_UNEXPECTED` followed by an `ob_abort`.

## Solution Description
Add `deserialize_without_buffer=true` to the VSAG `index_param` for SINDI. This bypasses the `BufferStreamReader` which depends on the full stream length, allowing SINDI to read the snapshot directly from the callback stream. The existing `deserialize_without_footer=true` is retained, ensuring no change to the snapshot format.
A new, independent mysqltest case is added to cover the scenario of executing an ANN query immediately after creating a SINDI index on a non-empty table.

## Passed Regressions
- `CARGO_NET_OFFLINE=true ob-make seekdb`: Passed.
- Executed regression SQL using the newly built seekdb: Both the first and subsequent ANN queries after index creation returned the expected results (2, 1).
- Restarted seekdb using the same data directory and queried again: Returned the expected results (2, 1).
- No occurrences of `BufferStreamReader`, `READ_ERROR`, `fdeserialize error`, or `-7604` were found in the logs.
- There is no standalone mysqltest executable locally; the full SQL from the new test case has been executed and passed on an actual seekdb instance.

## Upgrade Compatibility
No changes to the SINDI snapshot format; existing indexes remain compatible. Only the deserialization reading method has been adjusted.

## Other Information
This MR contains only one fix commit with no unrelated changes.

## Release Note
Fixed an issue where creating a local SINDI index on a non-empty table could cause the first ANN query to fail with error 7604 due to snapshot deserialization failure. The fix adjusts the deserialization process to correctly handle streamed data.

### DIFF
--- a/.akfarm
+++ b/.akfarm
@@ -7,7 +7,7 @@ ddlmigrate_lite
 ddltest_duomo
 multi_modal_test
 stress_duomo
-bigtrans_lite
+!bigtrans_lite
 obtrade_lite
 ps5_lite
 obrqg_lite


hnwyllmm:
The mapping Dima issue is [[seekdb][Vector Index] First ANN query after creating SINDI on a non-empty table fails with error 7604 due to snapshot deserialization failure.

footka (approved):

PR #1321, opened 2026-08-21
[compat] Fix 1.4 regressions: COM_STMT_EXECUTE zero-param compat byte rejected + ENGINE=MyISAM syntax error

footka:
## Release Note


footka:
The mapping Dima issue is [seekdb 1.4 protocol/syntax regression: COM_STMT_EXECUTE zero-parameter compatibility byte rejected + ENGINE=MyISAM syntax error]

hnwyllmm (approved):

PR #1322, opened 2026-08-21
[seekdb][DDL] Fix crash on ALTER TABLE ADD for generated timestamp column

hnwyllmm:
## Task Description
Fix a crash (SIGSEGV) and client error (ERROR 2013) that occurs in seekdb 1.4 when executing `ALTER TABLE ADD` to add a generated timestamp column.

## Solution Description
The regression was introduced by commit `2bf907c560fa` (remove oracle), which modified the short-circuit condition for timestamp columns. This inadvertently allowed generated columns to enter the ordinary timestamp default-value handling path, leading to the crash.

The fix skips the timestamp default-value resolution specifically for generated columns added via `ALTER TABLE`. This preserves the generated expression as the original default value while maintaining correct behavior for ordinary timestamp columns.

## Passed Regressions
- Built seekdb from the issue branch and executed the following test:
  ```sql
  CREATE TABLE t1 (c1 varchar(100), c2 timestamp AS (c1));
  ALTER TABLE t1 ADD COLUMN c3 timestamp AS (c1);
  ```
- Verified via `DESC` and `SHOW CREATE TABLE` that `c3` is correctly defined as a `timestamp VIRTUAL GENERATED` column.
- Confirmed the observer process remained reachable after the operation.
- A regression test for adding a normal timestamp column via `ALTER TABLE ADD` also passed.

## Upgrade Compatibility
No compatibility impact. The change only restores the correct handling for generated columns and does not alter the default value behavior of ordinary timestamp columns.

## Other Information
- The MR contains one focused commit.
- An untracked local file named `2` is unrelated and has been excluded from the changes.
- Related internal link: DIMA-2026082000118399861

## Release Note
Fixed a crash that could occur when using `ALTER TABLE ADD` to add a generated timestamp column.

### DIFF
--- a/src/rootserver/ob_ddl_service.cpp
+++ b/src/rootserver/ob_ddl_service.cpp
@@ -7110,6 +7110,8 @@ int ObDDLService::resolve_timestamp_column(AlterColumnSchema *alter_column_schem
     ret = OB_ERR_UNEXPECTED;
     LOG_WARN("alter_column_schema is NULL", K(ret));
   } else if (ObTimestampType != new_column_schema.get_data_type()
+      || (OB_DDL_ADD_COLUMN == alter_column_schema->alter_type_
+          && new_column_schema.is_generated_column())
       || new_column_schema.is_generated_column()
       == alter_column_schema->check_timestamp_column_order_) {
     //nothing to do


hnwyllmm:
The mapping Dima issue is [seekdb 1.4: ALTER TABLE ADD generated column (timestamp expression) causes observer SIGSEGV crash (regression from 1.3 to 1.4)]

footka (approved):

PR #1332, opened 2026-08-27
fix pkg

ep-12221:
## Release Note


ep-12221:
The mapping Dima issue is fixed in the package.

footka (approved):

PR #1334, opened 2026-08-27
Fix Windows start problem

ep-12221:
## Release Note


ep-12221:
The mapping Dima issue is about [windows startup failure]

PR #1338, opened 2026-08-28
Optimize Named Locks by Removing __all_dbms_lock_allocated Persistence

Mai-icy:
### Task Description

Migrate [oceanbase/oceanbase-lite!1089](https://gitlab.oceanbase-dev.com/oceanbase/oceanbase-lite/-/merge_requests/1089) to SeekDB.

The change removes persistence through `__all_dbms_lock_allocated` and keeps MySQL named-lock state in memory.

### Solution Description

- Keep MySQL named locks in the in-memory named-lock manager.
- Update session cleanup and reset-connection handling.
- Expose named locks and their object names through the virtual object-lock view.
- Preserve named-lock cleanup errors.
- Remove obsolete named-lock inner-table helpers and storage inventory entries.

### Passed Regressions

- `git diff --check upstream/master...HEAD`
- Full compilation and regression testing have not yet been run on the migrated branch.

The original MR updated mysqltest coverage. Five legacy mysqltest files no longer exist on the current GitHub `master`, so their changes were not reintroduced:

- `tools/deploy/mysql_test/r/mysql/simple_ddl.result`
- `tools/deploy/mysql_test/test_suite/table_lock/r/mysql/lock_func_basic.result`
- `tools/deploy/mysql_test/test_suite/table_lock/t/lock_func_basic.test`
- `tools/deploy/mysql_test/test_suite/view/r/mysql/cdb_objects.ce.result`
- `tools/deploy/mysql_test/test_suite/view/r/mysql/cdb_objects.result`

Updates to the two test-result files still present in SeekDB were migrated.

### Upgrade Compatibility

This changes named-lock state from inner-table persistence to in-memory runtime state. Upgrade and restart behavior should be reviewed before merge.

### Other Information

- Original GitLab MR: [oceanbase/oceanbase-lite!1089](https://gitlab.oceanbase-dev.com/oceanbase/oceanbase-lite/-/merge_requests/1089)
- Original source branch: `task/2026072800117800650`
- Original GitLab tip: `1d7f1b4bf05a287147da020de12049dfd249176f`
- The seven original commits were transplanted in order onto the latest `oceanbase/seekdb:master`; authorship and commit messages were preserved.
- GitLab and GitHub use unrelated Git histories, so commit SHAs changed during migration.

### Release Note

Manage MySQL named locks in memory, expose them through the virtual object-lock view, and remove obsolete `__all_dbms_lock_allocated` persistence helpers.


src/observer/mysql/obmp_reset_connection.cpp
```diff
@@ -169,11 +169,8 @@ int ObMPResetConnection::process()
     if (OB_SUCC(ret)) {
       const data_plane::ObSessionLockOwner owner(
           session->get_sid(), session->get_sess_create_time());
-      data_plane::ObPersistedLockOwner persisted_owner;
-      if (OB_FAIL(data_plane::persist_session_lock_owner(owner,
-                                                         persisted_owner))) {
-      } else if (OB_FAIL(query::release_locks_for_dead_owner(
-                     persisted_owner.owner_type_, persisted_owner.owner_id_))) {
+      int64_t release_count = 0;
+      if (OB_FAIL(data_plane::release_all_named_locks(owner, release_count))) {
```
wangyunlai-seekdb: **[P1] Preserve table-lock cleanup during connection reset** — The replaced `release_locks_for_dead_owner()` call went through the `ALL_LOCKS` path, which releases both object/named locks and session-level `LOCK TABLES` locks. This call only clears `NamedLockManager`. Because `COM_RESET_CONNECTION` keeps the same live session, `LOCK TABLES t WRITE; COM_RESET_CONNECTION` can leave `t` locked and block another session indefinitely. Keep the persisted table-lock cleanup, or explicitly release the table-lock scope, alongside named-lock cleanup and cover this with a two-session reset test.

src/storage/tablelock/ob_named_lock_manager.cpp
```diff
@@ -0,0 +1,455 @@
...
+  return ret;
+}
+
+bool NamedLockManager::would_deadlock_(const ObTableLockOwnerID &waiter,
+                                       const ObTableLockOwnerID &blocker) const
+{
+  bool deadlock = blocker == waiter;
+  ObTableLockOwnerID current = blocker;
+  for (int64_t depth = 0; !deadlock && depth <= static_cast<int64_t>(wait_for_map_.size()); ++depth) {
+    WaitForMap::const_iterator it = wait_for_map_.find(current);
```
wangyunlai-seekdb: **[P1] Integrate named-lock waits into the shared deadlock graph** — This detector follows only the manager-local `wait_for_map_`. Before this change, `MYSQL_LOCK_FUNC` and `LOCK TABLES` both went through `ObTableLockService` and its shared detector. A cycle can now be split across the two graphs: A owns named lock N, B owns table T, A waits for T, then B waits for N. Neither graph sees the full cycle, so both sessions wait until timeout instead of reporting a deadlock. Bridge named-lock edges into the existing detector and add an integration test for this mixed cycle.

unittest/storage/transaction/tablelock/test_named_lock_manager.cpp
```diff
@@ -0,0 +1,287 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <chrono>
+#include <cstdio>
+#include <thread>
+#include <vector>
+
+#include <gtest/gtest.h>
```
wangyunlai-seekdb: **[P1] Register the named-lock tests with a runnable target** — This file is not registered by a unit-test build target. Current CI only builds the `seekdb` binary and runs mysqltest slices, and the active mysqltests do not exercise `GET_LOCK`. Consequently, this concurrency, deadlock, cleanup, and quota suite is never compiled or executed. Add the appropriate unit-test target and CI invocation, or move equivalent coverage into an active mysqltest suite.

wangyunlai-seekdb (commented):
Three Major findings remain at the current PR head: connection reset can retain table locks, mixed named/table wait cycles escape deadlock detection, and the new test suite is not wired into a runnable target.

src/storage/tablelock/ob_named_lock_manager.cpp
Mai-icy: The old path did not reliably detect this mixed cycle either. Although both locks entered ObTableLockService, their acquisition used short-lived internal transactions. After the statement ended, the transaction detector nodedisappeared while the session lock remained held.
We intentionally use timeout-based resolution for named-lock wait cycles instead of deadlock detection. Named-lock waits are bounded by the smaller of the GET_LOCK timeout and query timeout (10 seconds by default), so mixed named/table cycles also terminate by timeout. Table-lock deadlock detection remains unchanged.

unittest/storage/transaction/tablelock/test_named_lock_manager.cpp
Mai-icy: Agreed—the file was not part of any runnable target and should not be treated as CI coverage. It has been removed in the latest revision together with the named-lock wait-for/deadlock logic. We are not adding a new unit-test target in this change.

hnwyllmm (approved):

PR #1352, opened 2026-09-01
fix: improve change stream consistency and eliminate idle statistics writes

hnwyllmm:
## Summary

This ports internal `oceanbase/oceanbase-lite!1067` onto the current GitHub `master`.

- separate Worker-committed Change Stream progress from the in-memory refresh proof watermark
- make explicit refresh schema-consistent across schema readiness, GTS capture, LSN fencing, commit barriers, and recovery epochs
- cache the Change Stream minimum dependency LSN in memory so checkpoint can advance without periodic internal-table writes while IDLE
- move automatic optimizer statistics maintenance to the existing 15-minute timer and skip internal-table scans and writes after a successful idle reconciliation
- filter statistics-maintenance internal tables from DML and column-usage feedback

## Change Stream

The persisted global state is now `change_stream_applied_scn`, advanced only after a Worker transaction commits. Dispatcher recovery and transaction skipping use only this committed fact.

`refresh_scn_` remains an in-memory API proof watermark. Explicit refresh waits for a stable runtime schema version, Fetcher readiness, a GTS target, an LSN fence, and a dispatched-work commit barrier. Deleting the final async index is handled through the schema drained watermark, while recovery invalidates stale in-flight proofs through an epoch.

Fetcher publishes `min_dep_lsn` through an atomic cache. ACTIVE mode keeps the existing monotonic persistence; IDLE mode updates only memory. Checkpoint reads the cached dependency boundary directly. Recovery additionally considers PALF base LSN and rejects hints beyond the current log end.

## Optimizer statistics

The built-in async gather job is marked `TIMER_DRIVEN`, so the generic Scheduler no longer executes it or updates its date/status fields. `ObOptStatMonitorCheckTask` is the single periodic trigger.

Monitoring flush, stale marking, invalid-row cleanup, and async gather use one serialized maintenance path. Epoch tracking preserves concurrent DML and enables a fast idle path after a complete reconciliation. Empty or filtered caches perform no SQL, and statistics-maintenance internal tables are excluded at both producer and flush boundaries.

Explicit procedure calls remain supported, while automatic heavy collection is dispatched off the shared timer thread.

## Public port notes

- retained the public dependency-injected Change Stream log-storage and schema-publish interfaces
- added `TIMER_DRIVEN` where the public tree currently defines `ObDBMSSchedJobInfo`
- adapted Scheduler availability checks to the public SQL/module layering
- did not copy the source MR's internal-only mysqltest and mock files, which are absent from this repository

## Validation

- `./build.sh release --init`
- full `ob-make` build in `build_release`, without `-j`
- `build_release/src/observer/seekdb --version`
- `git diff --check`

The public checkout currently exposes no matching runnable CTest/mysqltest cases (`ctest -N` reports 0 tests). The internal-only test files from the source MR therefore could not be migrated or run here.


src/observer/change_stream/ob_change_stream_mgr.cpp
```diff
@@ -131,39 +166,147 @@ int ObChangeStreamMgr::wait_refresh_scn(
...
+      ret = OB_TIMEOUT;
+    } else if (OB_FAIL(OB_TS_MGR.get_gts_sync(
+                   abs_timeout_us - ObTimeUtility::current_time(), target_scn))) {
+      LOG_WARN("get GTS for change stream refresh failed", KR(ret));
+    } else if (OB_FAIL(GCTX.schema_service_->get_runtime_refreshed_schema_version(
+                   schema_version_v2))) {
+      LOG_WARN("failed to get runtime schema version after refresh target", KR(ret));
+    } else if (schema_version_v1 != schema_version_v2) {
+      // The (schema state, GTS) pair was not captured from one stable schema.
+      continue;
```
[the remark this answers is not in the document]
hnwyllmm: Fixed in d3572ae59a55. The schema-version recapture path now sleeps for 1 ms before retrying, while continuing to use the original absolute deadline.

src/sql/optimizer/stat/ob_stat_define.h
```diff
@@ -16,3 +16,11 @@
 
 #pragma once
 #include "query/optimizer/stat/ob_stat_define.h"
+
+namespace oceanbase
+{
+namespace common
+{
+const static int64_t OPT_STATS_MAINTENANCE_INTERVAL_US = 15L * 60L * 1000L * 1000L;
```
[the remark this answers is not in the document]
hnwyllmm: Fixed in d3572ae59a55: `OPT_STATS_MAINTENANCE_INTERVAL_US` is now declared `static constexpr`.

src/storage/tx_storage/ob_checkpoint_service.cpp
```diff
@@ -19,8 +19,8 @@
 #include "storage/tx_storage/ob_checkpoint_service.h"
 #include "share/rc/ob_server_runtime.h"
 #include "logservice/ob_log_service.h"
-#include "share/ob_global_stat_proxy.h"
 #include "share/ob_server_struct.h"
+#include "observer/change_stream/ob_change_stream_mgr.h"
 #include "storage/tx_storage/ob_ls_service.h"
```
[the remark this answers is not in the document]
hnwyllmm: Fixed in d3572ae59a55. Checkpoint no longer includes or looks up the observer implementation. `get_min_dep_lsn()` is exposed through the existing narrow `query::ObIChangeStreamService`, registered via the server runtime, matching the existing active-snapshot service pattern.

src/observer/change_stream/ob_change_stream_dispatcher.cpp
```diff
@@ -140,27 +150,59 @@ void ObCSDispatcher::destroy()
...
+}
+
+void ObCSDispatcher::inc_epoch()
+{
+  common::ObSpinLockGuard guard(refresh_epoch_lock_);
+  ATOMIC_STORE(&recovery_in_progress_, 1);
+  ATOMIC_INC(&epoch_);
+}
+
+int ObCSDispatcher::publish_refresh_scn_if_epoch(
```
hnwyllmm: rename to try_publish_refresh_scn

src/observer/change_stream/ob_change_stream_dispatcher.cpp
```diff
@@ -140,27 +150,59 @@ void ObCSDispatcher::destroy()
...
+    const int64_t expected_epoch,
+    bool &published)
+{
+  int ret = OB_SUCCESS;
+  published = false;
+  common::ObSpinLockGuard guard(refresh_epoch_lock_);
+  if (ATOMIC_LOAD(&epoch_) == expected_epoch
+      && 0 == ATOMIC_LOAD(&recovery_in_progress_)) {
+    ObChangeStreamMgr *mgr =
+        ::oceanbase::share::server_service<::oceanbase::share::ObChangeStreamMgr>();
```
hnwyllmm: ::oceanbase::share::ObChangeStreamMgr --> ObChangeStreamMgr
::oceanbase::share::server_service --> share::server_service

src/observer/change_stream/ob_change_stream_fetcher.cpp
```diff
@@ -103,38 +109,103 @@ int ObCSFetcher::init_consumption_position_()
...
+  palf::LSN cached_min_dep_lsn;
+  storage::ObLS *ls = nullptr;
+  logservice::ObLogHandler *log_handler = nullptr;
   if (OB_ISNULL(GCTX.sql_proxy_)) {
     ret = OB_ERR_UNEXPECTED;
     LOG_WARN("CSFetcher: sql_proxy is null", KR(ret));
   } else if (OB_FAIL(ObGlobalStatProxy::get_change_stream_min_dep_lsn(
                  *GCTX.sql_proxy_, false, persisted_min_dep_lsn))) {
+    LOG_WARN("CSFetcher: fail to load change_stream_min_dep_lsn", KR(ret));
+  } else if (OB_FAIL(::oceanbase::share::server_service<::oceanbase::storage::ObLSService>()->get_ls(ls))) {
```
hnwyllmm: ::oceanbase::storage::ObLSService --> ObLSService or storage::ObLSService

src/observer/change_stream/ob_change_stream_fetcher.cpp
```diff
@@ -238,32 +320,64 @@ void ObCSTxInfo::destroy()
...
 {
   int ret = OB_SUCCESS;
   min_lsn.reset();
 
   palf::LSN end_lsn;
   storage::ObLS *ls = nullptr;
-  if (OB_FAIL(::oceanbase::share::server_service<::oceanbase::storage::ObLSService>()->get_ls(ls))
-      || OB_FAIL(ls->get_log_handler()->get_end_lsn(end_lsn))) {
+  logservice::ObLogHandler *log_handler = nullptr;
+  if (OB_FAIL(::oceanbase::share::server_service<::oceanbase::storage::ObLSService>()->get_ls(ls))) {
```
hnwyllmm: ::oceanbase::storage::ObLSService --> ObLSService

src/observer/change_stream/ob_change_stream_dispatcher.cpp
wangyunlai-seekdb: Fixed in 9031eb1f734b: renamed `publish_refresh_scn_if_epoch()` to `try_publish_refresh_scn()` in the declaration, definition, and both call sites. The epoch guard remains documented in the function contract.

src/observer/change_stream/ob_change_stream_dispatcher.cpp
wangyunlai-seekdb: Fixed in 9031eb1f734b: simplified the lookup to `ObChangeStreamMgr` and `server_service`. I also scanned every line introduced by this PR and removed the analogous unnecessary `::oceanbase::` qualification from the other affected files.

src/observer/change_stream/ob_change_stream_fetcher.cpp
wangyunlai-seekdb: Fixed in 9031eb1f734b: simplified the service lookup to `ObLSService`. The same cleanup was applied to the other newly introduced occurrences in this PR.

src/observer/change_stream/ob_change_stream_fetcher.cpp
wangyunlai-seekdb: Fixed in 9031eb1f734b: simplified this service lookup to `ObLSService` as well, together with all analogous overqualified lookups added by the PR.

src/observer/change_stream/ob_change_stream_fetcher.cpp
```diff
@@ -677,10 +927,10 @@ int ObCSFetcher::handle_commit_log_(
         ret = OB_ERR_UNEXPECTED;
         LOG_ERROR("CSFetcher: dispatcher_ is null", KR(ret));
       } else if (OB_FAIL(dispatcher_->push(tx))) {
-        LOG_WARN("CSFetcher: fail to push tx to dispatcher, destroy tx to avoid leak", KR(ret), K(tid));
-        (void)tx_info_.erase_refactored(tid);
-        tx->destroy();
-        OB_DELETE(ObCSTxInfo, "CSTxInfo", tx);
+        // Keep tx_info_ intact.  The iterator is rewound to current_lsn_ and
```
wangyunlai-seekdb, line 930: **[Major] Avoid replaying REDO retained after an enqueue failure**

A PALF block can contain REDO and COMMIT together. At this point, REDO has already been appended to `tx->redo_list_`; if `dispatcher_->push(tx)` fails, for example because `tx_ring_.set()` cannot extend the ring, retaining `tx` and returning an error makes `run1()` re-seek the unchanged `current_lsn_` at lines 1234–1241. The retry then appends the same REDO again before pushing. The original database transaction is already fully committed, while the later Worker `ObMySQLTransaction` has not started yet, so neither transaction can roll back this in-memory Fetcher mutation. The async-index path can therefore receive duplicate `(commit_version, vid, type)` rows; a failed Worker batch rolls back its SQL writes, but recovery retains the same duplicated ring entry and retries it, preventing `applied_scn` from advancing. Restore entry-local redo/rollback state before reseeking, or retry only `dispatcher_->push(tx)`. Please add a one-shot push-failure test using a compound REDO+COMMIT block.

wangyunlai-seekdb (commented):
I found one Major failure-path issue in the PALF-entry retry path; see the inline comment.

I also re-evaluated the optimizer-statistics epoch concern and intentionally did not report it as a Major finding. Lowering `ASYNC_GATHER_STALE_RATIO` after an epoch has completed can defer reconsideration of already-flushed DML when no new monitored DML arrives. However, any subsequent monitored DML advances the global epoch and causes another candidate scan on the next 15-minute maintenance pass; manual gathering, direct job invocation, or restart also recovers it. The impact is limited to delayed statistics refresh for tables newly crossing the lower threshold, without data-correctness or availability impact, so it does not meet the Major threshold.

src/observer/change_stream/ob_change_stream_fetcher.cpp
hnwyllmm: raise another issue: SEEK-532

hnwyllmm:
pipeline: 244194

PR #1354, opened 2026-09-01
Support random TCP ports in embedded mode

wangyunlai-seekdb:
## Summary

- derive TCP behavior directly from the SQL port argument: fixed port, OS-assigned port, or TCP disabled
- require the local Unix socket or named pipe when the port must be discovered
- expose the bound TCP port through one read-only NIO getter while keeping NIO ABI version 26 and the nio_start signature unchanged
- publish an atomic effective MySQL port after NIO starts and use it in mysql_port(), server statistics, and telemetry
- keep @@port unchanged

## Validation

- built the seekdb target from current upstream master
- Rust sql-nio tests: 7/7 passed, including fixed, ephemeral, and disabled TCP behavior
- Rust formatting passed
- git diff --check passed
- Linux runtime validation covered fixed server port, embedded loopback random port, effective-port SQL reporting, and same-directory single-instance behavior


hnwyllmm (dismissed):
LGTM

hnwyllmm:
CI  pipeline passed: 237523

src/observer/ob_server.cpp
```diff
@@ -1765,6 +1769,25 @@ int ObServer::init_opts_config(const ObServerOptions &opts, const char *optstr)
   if (opts.port_ != 0) {
     config_.mysql_port = opts.port_;
   }
+  if (opts.port_mode_ != ObServerOptions::PORT_MODE_DEFAULT) {
```
hnwyllmm: Why do you convert from string to enum in parse_line and convert enum to string here? Let's use string in both functions?

src/observer/ob_command_line_parser.cpp
```diff
@@ -505,6 +522,7 @@ void ObCommandLineParser::print_help() const
   MPRINT("Options:");
   MPRINT("  --nodaemon                      whether to not run as a daemon");
   MPRINT("  --port, -P <port>               the port, default is 2881");
+  MPRINT("  --port-mode <mode>              port mode: specified, random, or disabled");
```
hnwyllmm: Let's delete this parameter as the user wouldn't use it directly. We can also pass through the parameter by '--parameter'.

src/observer/ob_server.cpp
```diff
@@ -692,6 +692,11 @@ int ObServer::init(const ObServerOptions &opts, const ObPLogWriterCfg &log_cfg)
   }
   if (OB_SUCC(ret) && need_initialize) {
     LOG_INFO("Need to initialize", K(need_initialize));
+    if (gctx_.is_embedded_mode()
```
hnwyllmm: Let's the use or libseekdb decide what to do. So please delete the code.

src/observer/ob_srv_network_frame.cpp
```diff
@@ -80,11 +80,21 @@ void ObSrvNetworkFrame::destroy()
 int ObSrvNetworkFrame::start()
 {
   int ret = OB_SUCCESS;
-  const bool disable_tcp = gctx_.is_embedded_mode();
+  int mysql_port = static_cast<int>(GCONF.mysql_port);
+  const ObString port_mode = GCONF.mysql_port_mode.str();
+  if (port_mode == ObString("random")) {
```
hnwyllmm: We should use "case compare" but not case sensitive mode.

src/observer/ob_srv_network_frame.cpp
```diff
@@ -97,10 +107,14 @@ int ObSrvNetworkFrame::start()
...
     }
     if (OB_FAIL(obmysql::global_sql_nio_server->start(
-            GCONF.mysql_port, &deliver_, sql_net_thread_count,
-            disable_tcp, GCONF.ssl_client_authentication,
+            mysql_port, &deliver_, sql_net_thread_count,
+            GCONF.ssl_client_authentication,
             GCONF.sql_protocol_min_tls_version.str()))) {
-    } else if (OB_FAIL(reload_config())) {
+    } else {
+      gctx_.set_effective_mysql_port(
```
hnwyllmm: Please delete effective mysql port function and use global_sql_nio_server->get_bound_tcp_port directly.

src/share/parameter/ob_parameter_seed.ipp
```diff
@@ -56,8 +56,11 @@ DEF_PARAM(enable_rpc_tls, BOOL, OB_CLUSTER_PARAMETER, "False",
 DEF_PARAM(rpc_port, INT, OB_CLUSTER_PARAMETER, "2882", "(1024,65536)",
         "the port number for RPC protocol. Range: (1024, 65536) in integer",
         ObParameterAttr(Section::OBSERVER, Source::DEFAULT, EditLevel::DYNAMIC_EFFECTIVE));
-DEF_PARAM(mysql_port, INT, OB_CLUSTER_PARAMETER, "2881", "(1024,65536)",
-        "port number for mysql connection. Range: (1024, 65536) in integer",
+DEF_PARAM(mysql_port, INT, OB_CLUSTER_PARAMETER, "2881", "[0,65535]",
```
hnwyllmm: revert to [1024, 65535]

hnwyllmm (changes requested):

hnwyllmm (approved):

Issue #191, opened 2026-03-11
Support Android Compilation


LINxiansheng:

hnwyllmm:
This issue has been closed because the original work item content was empty. Empty content issue closed.

PR #1362, opened 2026-09-07
refactor(sql): manage SQL plan lifetime with session references

wangyunlai-seekdb:
### Task Description

The previous session-local plan-cache design duplicated physical plans across connections and caused unacceptable memory growth. Retain globally shared physical plans while making their lifetime follow active session usage.

### Solution Description

- Keep SQL plans shared in the tenant plan-cache lookup map and let each session retain strong references to the plans it uses.
- Bound each session to 100 retained plan references and replace the least recently used reference when full.
- Retire SQL plans synchronously after they are detached and the last session, execution, or diagnostic reference is released; exclude SQL plans from periodic idle eviction.
- Release session references on session teardown, RESET CONNECTION, and successful CHANGE USER.
- Preserve FLUSH and correctness eviction by detaching the old generation and lazily pruning stale session references.
- Keep periodic memory and idle maintenance for non-SQL library-cache objects.

### Passed Regressions

- `git diff --check`
- Built the SeekDB server target successfully.
- Lifecycle checks: two sessions share one physical plan; closing the last session releases it; FLUSH detaches the old generation and the next lookup creates a new one; one session retains at most 100 plans.
- 64-session, 32-PS point-select test with a 10-second workload warmup and 60-second measurement: 43,109.89 QPS versus 42,944.20 baseline (+0.39%), with no errors in the comparable adjacent run.
- A separate 10-second settle ABBA run measured +0.14% candidate versus baseline. Later shared-host rounds were excluded from aggregate conclusions because host load changed materially during the run.

### Upgrade Compatibility

No persistent data format, wire protocol, SQL syntax, or configuration changes. The change affects only in-memory SQL plan ownership and retirement.

### Other Information

Each session retains at most 100 shared plan references; multiple sessions using the same physical plan still share one plan object. The shared-node and object retirement concurrency path is the primary review focus.

### Release Note

Reduce SQL plan-cache memory growth by sharing physical plans across sessions and releasing plans when their last reference is gone.

### DIFF
--- a/src/observer/mysql/obmp_auth_response.cpp
+++ b/src/observer/mysql/obmp_auth_response.cpp
@@ -47,6 +47,8 @@ int ObMPAuthResponse::process()
   } else if (OB_FAIL(session->set_login_auth_data(auth_data_))) {
   } else if (OB_FAIL(load_privilege_info_for_change_user(session))) {
   } else {
+    // This is the successful completion of the auth-switch CHANGE USER path.
+    session->reset_session_plan_refs();
     conn->set_auth_phase();
     ObOKPParam ok_param; // use default values
     ok_param.is_on_change_user_ = true;
--- a/src/observer/mysql/obmp_change_user.cpp
+++ b/src/observer/mysql/obmp_change_user.cpp
@@ -88,6 +88,11 @@ int ObMPChangeUser::process()
       } else if (need_send_auth_switch) {
         // do nothing
       } else if (OB_FAIL(load_privilege_info_for_change_user(session))) {
+      } else {
+        // Privilege loading completes CHANGE USER for clients without an auth
+        // switch round trip. Drop the old user's plan references immediately;
+        // later PS/cursor cleanup failures must not retain the old identity.
+        session->reset_session_plan_refs();
       }
     }
   }
--- a/src/observer/mysql/obmp_reset_connection.cpp
+++ b/src/observer/mysql/obmp_reset_connection.cpp
@@ -192,6 +192,9 @@ int ObMPResetConnection::process()
 
 
     if (OB_SUCC(ret)) {
+      // RESET CONNECTION starts a fresh logical connection on the same
+      // session object; it must not inherit retained Plan references.
+      session->reset_session_plan_refs();
       session->clean_status();
     }
   }
--- a/src/observer/ob_server.cpp
+++ b/src/observer/ob_server.cpp
@@ -912,6 +912,15 @@ void ObServer::destroy()
 
   if (!has_destroy_ && has_stopped_) {
 
+    // Drain runtime workers and sessions before destroying the session manager
+    // or the plan-cache memory context. Sessions can retain SQL plan references
+    // between requests, so destroying the plan cache first would leave dangling
+    // references in sessions whose physical destruction is still deferred.
+    FLOG_INFO("begin to wait server runtime");
+    server_runtime_controller_.stop();
+    server_runtime_controller_.wait();
+    FLOG_INFO("server runtime wait finished");
+
     FLOG_INFO("begin to destroy OB_LOGGER");
     OB_LOGGER.destroy();
     FLOG_INFO("OB_LOGGER destroyed");
--- a/src/observer/ob_server.cpp
+++ b/src/observer/ob_server.cpp
@@ -912,15 +912,6 @@ void ObServer::destroy()
 
   if (!has_destroy_ && has_stopped_) {
 
-    // Drain runtime workers and sessions before destroying the session manager
-    // or the plan-cache memory context. Sessions can retain SQL plan references
-    // between requests, so destroying the plan cache first would leave dangling
-    // references in sessions whose physical destruction is still deferred.
-    FLOG_INFO("begin to wait server runtime");
-    server_runtime_controller_.stop();
-    server_runtime_controller_.wait();
-    FLOG_INFO("server runtime wait finished");
-
     FLOG_INFO("begin to destroy OB_LOGGER");
     OB_LOGGER.destroy();
     FLOG_INFO("OB_LOGGER destroyed");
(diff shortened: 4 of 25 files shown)


wangyunlai-seekdb:
## Performance test report: Jenkins #190 vs. #191

This comparison did not reproduce the severe performance degradation. TPS differences across the four 300-second workloads ranged from -0.70% to +1.51%. The two read-only workloads actually ran for only 10 seconds, with TPS decreases of 2.15% and 2.93%; longer, repeated runs are needed to determine whether these are consistent regressions. Sampled peak process RSS increased by 1.46%, with no order-of-magnitude memory growth observed. These measurements do not establish a reduction in Plan Cache-specific memory usage.

### Versions and configuration

| Item | PR version | Pre-PR baseline |
|---|---|---|
| Jenkins | [#190](http://11.166.86.153:9090/jenkins/job/lite_perf_guard_new/190/) | [#191](http://11.166.86.153:9090/jenkins/job/lite_perf_guard_new/191/) |
| Commit | `b21ea85034a23c4024f256ab259ae7ad26b85ec3` | `a48096008ca8a5869afc0a290c07ba6ebf3ac4b9` |
| Build mode | RelWithDebInfo | RelWithDebInfo |
| Jenkins status | SUCCESS | SUCCESS |

The baseline is the PR commit's direct parent. Both runs used the same target, `11.161.204.175:4345`, with 22 CPUs, 174 GiB of memory, 550 sysbench threads, 30 tables, and initially 100,000 rows per table. Both used `run_ecs.sh` SHA `611c1c436f7c412efc409cfde34d55207a86643a`. Random seeds and relevant settings in the executed commands matched. With `--db-ps-mode=disable`, coverage was limited to text SQL; prepared statements, TPC-H, TPC-DS, VectorDBBench, and MLDR were not tested.

Both builds selected `sysbench_runtime=300`, but `run_22c.sh` hard-coded `point_select` and `read_only` to 10 seconds per run. Each ran twice, and the reported result came from the second run. The other four workloads used 300 seconds. `read_only` and the four longer workloads also used `--warmup-time=1`. Therefore, this comparison must not be described as a 300-second test of every workload.

### Throughput and latency

Changes are calculated as `(PR / baseline - 1) × 100%`. Higher TPS and lower latency are better. Average latency and P99 are in milliseconds.

| Workload | Measured duration | Baseline TPS | PR TPS | TPS change | Average latency: baseline → PR | P99: baseline → PR |
|---|---:|---:|---:|---:|---:|---:|
| point_select | 10s | 235,717.11 | 230,642.99 | -2.15% | 2.29 → 2.34 | 9.39 → 10.65 |
| read_only | 10s | 10,162.58 | 9,865.25 | -2.93% | 53.91 → 55.03 | 262.64 → 272.27 |
| read_write | 300s | 5,718.92 | 5,680.53 | -0.67% | 96.17 → 96.82 | 189.93 → 193.38 |
| insert | 300s | 50,068.45 | 50,824.09 | +1.51% | 10.96 → 10.80 | 27.66 → 27.17 |
| update_non_index | 300s | 55,525.24 | 55,493.24 | -0.06% | 9.90 → 9.91 | 25.74 → 25.74 |
| write_only | 300s | 17,294.06 | 17,173.74 | -0.70% | 31.77 → 31.98 | 80.03 → 78.60 |

All runs reported zero `ignored errors` and zero `reconnects`. The point_select P99 increase of 1.26 ms (+13.42%) needs retesting. Only one complete test sequence was executed per version, which is insufficient to distinguish small performance differences from run-to-run variation.

### Memory observations

The following values are seekdb process RSS, sampled with a configured interval of 30 seconds. The first sample from the old process before each restart was excluded. The startup samples precede the 22-CPU/174-GiB configuration step in the logs and should not be treated as stable idle measurements after configuration.

| Observation | Baseline RSS (MiB) | PR RSS (MiB) | Change |
|---|---:|---:|---:|
| First sample after the new process started | 714.7 | 680.9 | -4.73% |
| Peak sampled RSS over the full run | 30,868.0 | 31,319.2 | +451.2 MiB (+1.46%) |
| Sample immediately after the workload ended | 12,024.2 | 13,163.3 | +1,139.1 MiB (+9.47%) |

Peak values were close, while the PR used approximately 1.11 GiB more at the end. RSS fluctuated by several GiB in both runs, and sampling times were not perfectly aligned, so the final sample alone cannot establish a leak. There was no 5–10-minute idle observation period after the workload, and no collection of Plan Cache allocator memory, cached plan counts, or hit rates. The RSS difference therefore cannot be directly attributed to Plan Cache.

### Validity and limitations

- Results were extracted from each build's raw console and cross-checked against the #190/#191 entries in `sysbench_trace.log`. The Jenkins HTML summary incorrectly associated old data with the new build and cannot be used directly for this comparison. Binary version extraction on the Jenkins node failed because required GLIBC versions were unavailable, leaving `build_info` empty. Startup logs from the actual target confirm the correct revisions above, and all sysbench workloads completed.
- Historical regression percentages in parentheses in the trace use the historical guard baseline, not this PR's parent commit. This report recalculates the differences between these two runs.
- Historical `sar` records covering the test period showed zero CPU steal, zero swap usage, and approximately 0.08% average iowait. These coarse measurements show no obvious steal, swap, or I/O-wait anomaly, but cannot rule out all background activity or explain differences of 2%–3%.
- The evidence supports “no severe performance degradation reproduced in this comparison,” rather than full performance acceptance. Follow-up testing should make both read-only workloads honor the configured 300-second duration, repeat A/B runs with identical warmup conditions, and collect post-workload idle RSS and Plan Cache-specific memory usage.

Raw evidence: [PR #190 console](http://11.166.86.153:9090/jenkins/job/lite_perf_guard_new/190/consoleText), [baseline #191 console](http://11.166.86.153:9090/jenkins/job/lite_perf_guard_new/191/consoleText).

hnwyllmm:
pipeline: 244856

PR #1388, opened 2026-09-14
fix(storage): preserve MINI merge progress under mixed compaction load

willlling:
Preserve MINI merge progress under mixed compaction load. The scheduler alternates MINI/MDS work when capacity is constrained, prefers large MINI work without starving FIFO work, and records fairness state only after successful dispatch.

This PR contains only the scheduler changes and its extracted-function regression harness (seven files). Optimizer warning handling and IN predicate changes are maintained independently in #1392 and #1393.

Local validation: the checked-in harness passed all 22 cases in both optimized and ASAN/UBSAN builds. It extracts the actual selection and dispatch-recording functions; it does not model full kernel concurrency.

Based directly on master `5441f5cc784c314ceefe3fc93d27066d55d58add` at head `4f8644917d16cec02a6b5e634f7ad34a5deebded`; no other PR in the six-PR set is a prerequisite. All 63 nonempty subsets were checked with 192 merge transitions: all 720 complete merge orders are conflict-free and converge to the same tree. `git diff --check` passed.

Current validation: [this PR’s SeekDB CI](https://github.com/oceanbase/seekdb/actions/runs/35299861120) and [the six-PR integration CI](https://github.com/oceanbase/seekdb/actions/runs/35299911515). These runs compile the server and execute the four mysqltest slices.

Restructuring and review: [SEEK-618](https://antmultica.alipay.com/seekdb/issues/SEEK-618).


tools/tests/mini_scheduler/main.cpp
footka: 这些测试不要入库

PR #1410, opened 2026-09-18
fix: avoid change stream refresh self-blocking on open transactions

cms-cms:
## Summary

- let the ChangeStream watermark advance past open transactions only when their redo is proven not to contain user-table DML
- keep unresolved user DML and committed/dispatched transactions behind the watermark
- prevent `FORK TABLE` / `FORK DATABASE` from waiting on their own table-lock or metadata-only transaction

## Root cause

`FORK TABLE` locks the source table and then waits for ChangeStream to reach a freshly sampled GTS before taking the snapshot. The lock transaction remains open during that wait.

The ChangeStream fetcher previously treated every entry in `tx_info_` as a watermark blocker. That includes the fork's own table-lock or metadata-only redo. The transaction cannot commit until the fork finishes, while the fork cannot finish until the watermark advances, creating a self-deadlock that ends as MySQL 4012.

The first fix was too broad because `commit_version_ == 0` only proves that Fetcher has not consumed the commit log. It does not prove that the transaction service has not assigned a commit SCN. Advancing past unresolved user DML could therefore make Dispatcher later skip an async-index update whose commit version is already at or below the watermark.

This revision classifies each redo buffer conservatively. Malformed or ambiguous redo is treated as user DML. Only open transactions whose redo is proven to contain no user-table DML are ignored by the refresh-watermark blocker. Unresolved user DML and every committed/dispatched transaction continue to block. Log retention remains unchanged because `get_min_dep_lsn()` still considers every `tx_info_` entry.

## Evidence

- Internal GitLab Pipeline 263937 / Job 1774693 reproduced MySQL 4012 in `test_fork_success[embedded]` at the original roughly 30-second wait boundary.
- macOS Job 1771578 reproduced the same fork failure at the same boundary.
- Validation Pipeline 264681 / Job 1784827 built the timeout-only patch exactly, but the three V2 fork cases still failed after the full roughly 1000-second DDL deadline.
- Validation Pipeline 264807 / downstream 264827 built the initial ChangeStream fix exactly; the complete PySeekDB embedded integration passed, including all fork cases.

## Validation

- Broad regression: internal GitLab Pipeline 265340 / downstream 265397 / Job 1796456 passed with `112 passed, 305 skipped, 878 deselected, 1 xpassed`; V1/V2 fork cases and the FTS scan/freeze regression all passed.
- Focused regression: internal GitLab Pipeline 266008 / downstream 266010 / Job 1806347 passed on the exact PR head `c60ac64913009d393c6434bc04e8284530b83e12`.
- The focused `vector_index.change_stream_refresh_scn_pending_dml` case uses a native asynchronous HNSW index and forces small-row redo before commit. It verifies that unresolved user DML is present, the refresh SCN does not advance while that transaction remains open, and after commit the pending count returns to zero and the inserted row is visible.
- The focused case passed on its first attempt in 2.619 seconds with no reboot or retry.
- The build used seekdb_test revision `20e8ede63222f8b7f3b84a4199e1d16104a035ca`; the resulting SeekDB package SHA-256 was `7f8de2a4e090b2fd49bc8f663b6ff7d1b70c66de2dfe733ab20bf1a792610dce`.


cms-cms:
将 PR 恢复为 Draft：外部 PySeekDB 源码构建验证中，3 个 V2 fork 专项里仍有 2 个在约 1000 秒后返回 4012（test_fork_success[embedded]、test_fork_preserves_original_collection[embedded]）。现有内核 CI 通过不足以证明该修复完成；待专项用例在源码 wheel 上通过后再标记 Ready。FTS/4016 验证与该 fork 修复分开处理。

PR #1297, opened 2026-08-12
[seekdb][tx] Keep commit retry task after accepted request

footka:
## Task Description
The issue was introduced by commit `8e2b24ff01af`, which stopped the commit retry scheduler from waiting for the final asynchronous callback after an initial commit request was accepted.

## Other Information
- Related links:
  - [DIMA-2026081100118160123](https://project.alipay.com/workItem?workItemId=2026081100118160123)

## Release Note

PR #1420, opened 2026-09-19
fix(tx): release outer latches before commit callbacks

willlling:
`CtxLockGuard::reset()` invokes commit callbacks through `unlock_ctx()` while still holding the access and redo latches. A concurrent commit retry can hold the transaction descriptor lock while waiting for one of those latches, forming a lock cycle. Release access and redo before unlocking the context, matching the existing `CtxLock::unlock()` order.

Validation on current master `6aa8d1548a3136d37017121401f5fab4ecd9bea0`:

- A deterministic C++ concurrency fixture compiles the actual production reset method. The original method fails forward progress in five of six guard modes; the patch passes all six, including callback-once and idempotent reset checks. This is a focused latch/callback fixture, not a full-server concurrency test.
- The release build at `b8400dc9dc460746ed9c161c0ba2893f3290f712` passes four complete mysqltest cases with repository CI initialization and no retry: commit-after-minor-merge, rollback, implicit commit, and set-transaction.
- The original BigTrans workload was also run with this patch plus the two rollback fixes from #1415, using its original JAR/configuration, 6 CPUs, 30G memory and 300-second pressure duration. It progressed beyond the previously observed initialization stall and all 16 duration timers completed. **The workload still failed** because PdmlTransaction reported a transaction timeout; this is not a claim of an overall BigTrans PASS. The prior rollback-only build stalled during initialization and reached the 900-second guard.

The bad release order already exists in public initial import `2c524b5424cb77d63058d166d088e9cd90a99ff0`; that method also fails the same fixture. The import author and committer are wangzelin.wzl, which does not establish authorship of the original defect. #1297 (author footka, committer GitHub, merged by footka) is a candidate for exposing the cycle through retained retry tasks, but no historical full-server first-bad/bisect is claimed.

Daily investigation: [SEEK-634](https://antmultica.alipay.com/seekdb/issues/SEEK-634). Original daily regression remains FAILED.


hnwyllmm:
@codex review

footka (approved):

Issue #264, opened 2026-03-12
Update obshell 4.4.1.1 for compatibility with seekdb 1.2.0


hnwyllmm:
## Tenant Mode
Please specify the tenant mode(s) in which this requirement must be met. List all if applicable. (Required)

## Description
Please describe the user requirement here. Provide as much context, detail, and background information as possible. (Required)

## Impact
Describe the business impact if this requirement is not met, and include the customer's project timeline. (Required)

## Proposed Solution
This is an optional field for proposing a solution approach.

PR #1423, opened 2026-09-20
fix: separate block-stat memtable scan boundary

cms-cms:
## Summary

- keep the physical multi-version scan boundary for SSTable index scans
- derive a separate schema-rowkey boundary for memtable scans
- prepare that memtable boundary during initial iterator setup and every tablet refresh

## Root cause

The fix from #1384 only normalized the refresh boundary when `curr_endkey_` was already available. During the first on-demand refresh, `curr_endkey_` can still be null, so the newly constructed memtable iterator receives the original four-column physical SSTable start key while its schema contains two logical rowkey columns. `ObMemtableKey::encode()` then reports a BIGINT/BIGINT UNSIGNED mismatch and returns `-4016`.

The previous implementation also reused the normalized range for SSTable scans. This change keeps the two ranges independent so fixing the memtable key cannot move or broaden the SSTable block boundary.

## Evidence

Reproduced on macOS from current master `6aa8d1548a3136d37017121401f5fab4ecd9bea0`. The server trace records `ob_memtable_key.h` data/schema mismatch, followed by `ob_memtable_iterator`, `ob_block_stat_iter`, FTS block-max evaluation, and client-visible `4016`.

## Validation

- `git diff --check`
- Linux and macOS source-build regression pending; PR remains Draft until both complete.


cms-cms:
Cross-platform regression validation update (PR remains Draft):

- The combined validation source `bafb8f3b942805b913de6c1f4b7c13fa5e7a8785` includes this PR together with the prerequisite validation changes.
- Linux Pipeline 268704 / child 268706 / Job 1841969: both embedded FTS regression modes (`first-pass` and concurrent `scan-freeze`) passed; no 4016 appeared in the trace. The full suite completed successfully (`119 passed, 298 skipped, 878 deselected, 1 xpassed`).
- macOS Pipeline 268527 / child 268528: both embedded FTS regression modes also passed with no 4016.
- Environment/cache failures were fixed only in the downstream seekdb_test repository and were not mixed into this product PR.

The remaining macOS 4012 timeout is in the separate collection-fork path and is not evidence against this memtable scan-boundary fix.

cms-cms:
Follow-up refinement pushed as `f8bed6bc`:

- keep `shrink_scan_range()` limited to preserving the physical SSTable boundary;
- rebuild the logical memtable boundary only after tablet/read-info refresh;
- reject a non-sentinel boundary with fewer datums than the schema rowkey instead of passing an invalid range to memtable scan.

Cross-platform validation is running against this exact SeekDB SHA and PySeekDB PR #264 SHA `8bdee2a6`:

- Linux: https://gitlab.oceanbase-dev.com/obqa/seekdb_test/-/pipelines/268851
- macOS: https://gitlab.oceanbase-dev.com/obqa/seekdb_test/-/pipelines/268852

Each pipeline selects only `hybrid_vector_index_lite,test_pyseekdb` plus required build jobs, using the 30-minute profile.

PR #1426, opened 2026-09-21
refactor(sql-nio): cbindgen-generated nio.h and named host shims

cao1629:
Squash of the mysql-nio-v3 series, rebased onto master 6aa8d1548:

- generate nio.h with cbindgen from the sql-nio crate. The ABI version handshake (NIO_ABI_VERSION, nio_start's abi_version, NIO_START_EABI), the two-sided layout assertions (abi_layout.rs, ob_nio_abi_check.cpp) and the NIO_CLIENT_* defines go away; rust-checks fails when the committed header is stale.
- use one struct name on both sides: the C++ consumers use the CamelCase Nio* names, the cbindgen rename table is gone, Reactor is NioReactor.
- drop the to_string injected into the view structs; ob_iarray.h detects a to_string member instead, so the ABI structs stay plain C.
- replace the NioCallbacks vtable with four named extern "C" shims, ob_sql_sock_handler_on_{connect,readable,disconnect,close}, that take the handler as their first parameter; nio_start takes the ObSqlSockHandler* directly instead of the table and its size.
- drop the cfg(test) stubs for the shims and the unreferenced ObClientCapabilityPos.

Master's additions to the hand-written header (nio_tls_string_view, nio_tls_session_info, nio_get_tls_session_info, min_tls_version, nio_get_bound_tcp_port, the NIO_TLS_MIN_* values) are carried by the generated header. The NIO_TLS_MIN_* constants live at the crate root: the four that only C++ reads would otherwise be flagged as dead code.


hnwyllmm:
@codex review

src/oblib/lib/container/ob_array.h
```diff
@@ -742,25 +742,29 @@ template<typename T, int64_t LOCAL_ARRAY_SIZE, typename BlockAllocatorT, bool au
...
-                                 : OB_LOG_KEEP_SIZE);
   J_ARRAY_START();
-  //delete now.If needed, must add satisfy json and run passed observer pretest
-  //common::databuff_printf(buf, buf_len, pos, "%ld:", count_);
-  for (int64_t index = 0; (index < need_print_count - 1) && (pos < buf_len - 1); ++index) {
-    if (pos + log_keep_size >= buf_len - 1) {
-      BUF_PRINTF(OB_LOG_ELLIPSIS);
+  if constexpr (has_to_string<T>::value) {
+    int64_t need_print_count = (count_ > max_print_count_ ? max_print_count_ : count_);
+    const int64_t log_keep_size = (ObLogger::get_logger().is_in_async_logging()
```
footka: This only controls whether or not to print; it should not modify the details.

footka (approved):

PR #1429, opened 2026-09-22
fix(observer): report inconsistent startup state before logger init

AbyssWhalen:
### Task Description

When `block_file` exists but the redo directory is empty, seekdb correctly refuses to start, but the reason is missing from redirected startup output and `seekdb.log`.

`check_need_initialize` runs before `OB_LOGGER.init`. Its `LOG_WARN` calls cannot allocate a log item at that point, and the `LOG_STDERR` fallback only writes to a TTY. Raising the log level to DEBUG does not recover the diagnostic.

### Solution Description

Write the existing reason, error code and base/data/redo paths directly to stderr, then flush before startup exits. This also reaches `seekdb.log` when startup has redirected stderr there. The error return, storage checks and initialization order stay unchanged; the patch does not repair or remove data.

The production change is intentionally limited to the startup diagnostic path. A standalone Linux regression was used locally to verify WARN and DEBUG startup with redirected output, the exact diagnostic, exit status 1, and an unchanged incomplete fixture; it is not included in this PR at the maintainer's request.

### Passed Regressions

Validated on Ubuntu 24.04 / WSL2 against `35eb2c3a3eafd9afff6457086e3141a83a52a6bd`:

- Release builds succeeded before and after the change.
- The same local regression on unmodified code: 11 log-reader checks pass; both startup-diagnostic checks fail because the message is absent.
- With the fix: all 13 local checks pass.
- A fresh instance starts, returns `SELECT VERSION(), 1`, shuts down cleanly, and repeats the query after restart.
- Python/Bash syntax checks and `git diff --check` pass.

The public checkout does not provide `unittest/CMakeLists.txt` or the documented `observer_tests` target. Private Farm, the full mysqltest suite, and Windows/macOS execution were not run. The standalone local regression is not part of this PR.

### Upgrade Compatibility

No schema, persisted-format or configuration changes. Inconsistent storage is still rejected.

### Other Information

AI-assisted; the patch was reviewed and tested locally.

### Release Note

Show the reason and directory paths when startup rejects inconsistent block-file/redo state, including redirected startup.


hnwyllmm:
Thanks for your contribution. Please delete the test file: unittest/observer/test_startup_diagnostics.py and then we can merge the pull request.

AbyssWhalen:
Thanks for the guidance. I removed `unittest/observer/test_startup_diagnostics.py` in commit `ded41484` and updated the PR description to clarify that the regression was run locally but is not included in this PR. Please take another look.

